PR #2, opened 2016-09-22
Add a role for resource managers

AndrewVos:
At the moment this just points to the root path.

Adds capybara, gds-sso and rspec.


spec/support/factory_girl.rb
```diff
@@ -0,0 +1,4 @@
+
```
benlovell: Did rubocop not complain about this line?

AndrewVos: Weirdly no

PR #3, opened 2016-09-22
Add govuk_admin_template

AndrewVos:


app/views/layouts/application.html.erb
```diff
@@ -1,14 +1,6 @@
-<!DOCTYPE html>
-<html>
-  <head>
-    <title>TelephoneAppointmentPlanner</title>
-    <%= csrf_meta_tags %>
+<% content_for :head do %>
+  <%= stylesheet_link_tag "application", media: "all" %>
```
benlovell: Nitpick: can we stick to single-quoted strings in ruby code (be that templates or otherwise)? Obviously rubocop doesn't lint ERB files so we have to keep on top of it this way unfortunately.

AndrewVos: Done

PR #7, opened 2016-09-22
Adding the ability to write es6

Guntrisoft:
- adding package.json to install babel and eslint
- running npm install on travis
- enable sprockets-es6
- adding basic bower.json

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,7 @@
+{
+  "name": "telephone_appointment_planner",
+  "private": true,
+  "dependencies": {
+
+  }
+}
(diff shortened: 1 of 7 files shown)


Guntrisoft:
@benbarnett

bower.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "telephone_appointment_planner",
+  "private": true,
+  "dependencies": {
+
```
benlovell: hurrrrr

benbarnett: can you stop breathing on pull requests please

Guntrisoft: I was in two minds on whether to add the bower stuff at this point when we don't have any dependencies at this moment, but we will eventually have full calendar, list.js etc etc

AndrewVos:
Merge it!

PR #8, opened 2016-09-22
Notify on build status in slack

AndrewVos:


.travis.yml
```diff
@@ -8,3 +8,5 @@ node_js:
 - '0.10'
 before_script:
 - bundle exec rake db:create db:migrate RAILS_ENV=test
+notifications:
+  slack: pensionwise:zpX4Fvisb1njvqag60SkbLkI
```
benlovell: Is that encrypted? 😱

PR #11, opened 2016-09-23
Add simple UI for editing users available slots

AndrewVos:


spec/features/shift_patterns_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.feature 'shift patterns' do
```
benlovell: Now these feature specs are becoming more meaningful we should adopt the same style as the planner project's features. I.e expressing the steps in the given-when-then style.

spec/models/user_spec.rb
```diff
@@ -4,4 +4,16 @@
 
 RSpec.describe User, type: :model do
   it_behaves_like 'a gds-sso user class'
+
+  describe '#resource_manager?' do
+    it 'returns false' do
+      user = build(:user)
+      expect(user.resource_manager?).to be false
```
benlovell: Maybe we should use the predicate matchers instead so these would read like `expect(user).to be_resource_manager` wdyt?

spec/models/user_spec.rb
```diff
@@ -4,4 +4,16 @@
...
+    end
+
+    it 'returns true if the user is a resource_manager' do
```
benlovell: If we mention 'if' in the spec text we should probably extract a context.

app/models/user.rb
```diff
@@ -2,4 +2,16 @@ class User < ApplicationRecord
   include GDS::SSO::User
   serialize :permissions, Array
   RESOURCE_MANAGER_PERMISSION = 'resource_manager'.freeze
+  GUIDER_PERMISSION = 'guider'.freeze
+
+  def self.guiders
+    # This can't really be made faster because we're storing
+    # permissions as a serialized string. We only really should have a maximum of
+    # 50 (guiders) + 3 (resource managers) in the database though.
+    User.all.select { |u| u.permissions.include? GUIDER_PERMISSION }
```
benlovell: If we expose a `guider?` predicate like you've done for resource managers we could golf this down to `User.all.select(&:guider?)`

app/views/users/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h2>Manage Guiders</h2>
+<% User.guiders.each do |guider| %>
```
benlovell: Nitpick: Can we assign the guiders to an ivar in the controller instead? Just so the controller is explicit about what we reach into with the views.

spec/features/roles_spec.rb
```diff
@@ -28,8 +28,10 @@
     end
 
     scenario 'allows resource managers to manage resources' do
-      visit '/'
-      expect(page).to have_content 'Welcome'
+      visit users_path
```
benlovell: We're going to see the benefits of page objects now these are getting more meaningful. Perhaps we should introduce SitePrism?

PR #12, opened 2016-09-26
Rotate travis/slack keys

AndrewVos:


benlovell:
This one only just landed in master so I don't think it needs rotating.

PR #13, opened 2016-09-28
Favour seeds over latent user creation

benlovell:
This causes problems when migrating on a freshly minted database.
GDS-SSO's initializer is loaded thus the `User` creation is attempted
and fails since the schema is not yet in place.


Guntrisoft:
👍 🚢

PR #16, opened 2016-09-29
Delete tags

benlovell:
Gives the resource manager the ability to delete tags.

### DIFF
--- a/app/views/tags/index.html.erb
+++ b/app/views/tags/index.html.erb
@@ -16,12 +16,19 @@
         <thead>
           <tr class="table-header">
             <th>Name</th>
+            <th></th>
           </tr>
         </thead>
         <tbody>
           <% @tags.each do |tag| %>
             <tr class="t-tag">
               <td><%= tag.name %></td>
+              <td><%= link_to 'Delete',
+                    tag, method: :delete,
+                    class: 'btn btn-danger t-delete',
+                    data: { confirm: "Are you sure you want to delete the tag #{tag.name}?" }
+                  %>
+              </td>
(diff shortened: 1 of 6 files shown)


app/views/tags/index.html.erb
```diff
@@ -16,12 +16,19 @@
...
           <tr class="table-header">
             <th>Name</th>
+            <th></th>
           </tr>
         </thead>
         <tbody>
           <% @tags.each do |tag| %>
             <tr class="t-tag">
               <td><%= tag.name %></td>
+              <td><%= link_to 'Delete',
```
benbarnett: Just had a quick chat here about 'tags' vs 'groups'. The users of this system have been shown a number of iterations of the prototype that refer to this functionality as groups...

Could we stick with that for anything in the UI? I guess it's fine to have them described as tags in the model, but their understanding on the domain language is a 'group' for this kind of thing.

benlovell: That's cool with me. Although I'm going to go the whole 🐷 and rename them everywhere.

app/views/tags/index.html.erb
```diff
@@ -16,12 +16,19 @@
         <thead>
           <tr class="table-header">
             <th>Name</th>
+            <th></th>
```
Guntrisoft: I've been putting 'Actions' in here as the header - it's wise to have a header for accessibility even if you wrap it in span with a sr-only class to only show it to screen readers.

benlovell:
This will be superseded by an incoming PR.

PR #18, opened 2016-09-29
Add better display_title for Slot Ranges

AndrewVos:


app/models/slot_range.rb
```diff
@@ -2,4 +2,8 @@ class SlotRange < ApplicationRecord
   belongs_to :user
   has_many :slots, inverse_of: :slot_range, dependent: :destroy
   accepts_nested_attributes_for :slots, allow_destroy: true
+
+  def display_title
```
benlovell: This feels like a presenter concern. What do you think?

app/views/users/edit.html.erb
```diff
@@ -25,8 +25,8 @@
   <tbody>
   <% @guider.slot_ranges.each do |slot_range| %>
     <tr>
-      <td><%= "#{slot_range.from} ➡ onwards" %></td>
-      <td><%= link_to edit_user_slot_range_path(@guider, slot_range), title: "Edit Slot Range #{slot_range.from.to_s}", :class => 'btn btn-info' do %><span class="glyphicon glyphicon-pencil"></span><span class="sr-only">Edit Slot Range</span><% end %></td>
+      <td><%= "#{slot_range.display_title}" %></td>
+      <td><%= link_to edit_user_slot_range_path(@guider, slot_range), title: "Edit Slot Range #{slot_range.display_title}", :class => 'btn btn-info' do %><span class="glyphicon glyphicon-pencil"></span><span class="sr-only">Edit Slot Range</span><% end %></td>
```
benbarnett: Does it make sense to rename 'slot range' to 'schedules' at this point?

spec/features/slot_ranges_spec.rb
```diff
@@ -69,15 +69,17 @@ def when_they_save_the_users_time_slots
   end
 
   def then_they_are_told_that_the_guider_has_been_created
```
benlovell: This is now at odds with what is actually happening.

spec/features/slot_ranges_spec.rb
```diff
@@ -69,15 +69,17 @@ def when_they_save_the_users_time_slots
...
-    expect(page).to have_content "Guider \"#{@guider.name}\" has been created"
+    slot_range = @guider.slot_ranges.first
+    expect(page).to have_content "Slot Range \"#{slot_range.display_title}\" has been created"
   end
 
   def then_they_are_told_that_the_guider_has_been_updated
```
benlovell: Same.

app/models/slot_range.rb
```diff
@@ -2,4 +2,8 @@ class SlotRange < ApplicationRecord
...
+    "#{from.strftime('%d %B %Y')} ➡ onwards"
```
benbarnett: From the meetings yesterday it sounds like this will be simply to give them all an end date of `created_at + 1 year`. Wdyt?

spec/features/slot_ranges_spec.rb
```diff
@@ -69,15 +69,17 @@ def when_they_save_the_users_time_slots
   end
 
   def then_they_are_told_that_the_guider_has_been_created
-    expect(page).to have_content "Guider \"#{@guider.name}\" has been created"
+    slot_range = @guider.slot_ranges.first
+    expect(page).to have_content "Slot Range \"#{slot_range.display_title}\" has been created"
```
benlovell: I think this is being over-specific. Can we just use a page object and check for the presence of an element instead? This is stuff that changes so these style of matchers end up being super-brittle.

app/controllers/slot_ranges_controller.rb
```diff
@@ -18,13 +18,19 @@ def update
       @slot_range.slots.destroy_all
       @slot_range.update(slot_range_parameters)
     end
-    redirect_to edit_user_path(@guider), flash: { success: "Guider \"#{@guider.name}\" has been updated" }
+    redirect_to(
+      edit_user_path(@guider),
+      flash: { success: "Slot Range \"#{@slot_range.display_title}\" has been updated" }
```
benlovell: Put this in the `ApplicationController` - `add_flash_types :success` and then you can:

```ruby
redirect_to edit_user_path(@guider), success: "Slot Range \"#{@slot_range.display_title}\" has been updated" }
```

spec/features/slot_ranges_spec.rb
AndrewVos: Fixed

spec/features/slot_ranges_spec.rb
AndrewVos: Fixed

app/views/users/edit.html.erb
AndrewVos: It's in my next task and would prefer to have it in a separate commit :)

app/models/slot_range.rb
AndrewVos: Don't think we need that quite yet because this UI could change a bit... Later on I was going to change onwards to display "whatever the next slot range from is"... except for the last one.

app/controllers/slot_ranges_controller.rb
AndrewVos: Nice!

app/views/users/edit.html.erb
benbarnett: 👍

app/controllers/slot_ranges_controller.rb
```diff
@@ -18,13 +18,19 @@ def update
       @slot_range.slots.destroy_all
       @slot_range.update(slot_range_parameters)
     end
-    redirect_to edit_user_path(@guider), flash: { success: "Guider \"#{@guider.name}\" has been updated" }
+    redirect_to(
+      edit_user_path(@guider),
+      success: "Slot Range \"#{@slot_range.display_title}\" has been updated"
```
benlovell: @benbarnett Do you think it's worth being so specific in these flash messages? I mean - referring to the actual thing that changed?

AndrewVos: Yeah I doubt this too... I mean it's not like they're going to look at the message and ensure that it's the correct slot range are they?

AndrewVos: We could even highlight the recently updated slot range in green?

benbarnett: Hmm... the thing is that schedules/slot ranges aren't actually editable. Once they're active, they've active. This is because they may have appointments allocated to their defined slots.

So the design was that they can create a new schedule any point from 6 weeks in advance. 

I suppose we could allow future schedules to be edited, but this feels like extra  logic at this stage?

benbarnett: Maybe it is simpler actually just to only allow editing if the schedule is due to 'start' more than 6 weeks in advance. So for example if a schedule has been created for 3 months time, then that's fine to edit.

benlovell: Makes sense to me.

app/models/slot_range.rb
AndrewVos: Done

app/controllers/slot_ranges_controller.rb
AndrewVos: Nice... I'll do this in a separate PR though!

app/presenters/slot_range_presenter.rb
```diff
@@ -0,0 +1,9 @@
+class SlotRangePresenter < SimpleDelegator
```
benlovell: ❤️

app/controllers/slot_ranges_controller.rb
```diff
@@ -18,13 +18,19 @@ def update
       @slot_range.slots.destroy_all
       @slot_range.update(slot_range_parameters)
     end
-    redirect_to edit_user_path(@guider), flash: { success: "Guider \"#{@guider.name}\" has been updated" }
+    redirect_to(
+      edit_user_path(@guider),
+      success: "Slot Range \"#{SlotRangePresenter.new(@slot_range)}\" has been updated"
```
benlovell: Can we just use 'The slot range has been updated' then or similar. i.e. don't refer to any of the actual slot range particulars?

app/controllers/slot_ranges_controller.rb
```diff
@@ -18,13 +18,19 @@ def update
...
+    )
   end
 
   def create
     @slot_range = @guider.slot_ranges.build(slot_range_parameters)
     @slot_range.save!
-    redirect_to edit_user_path(@guider), flash: { success: "Guider \"#{@guider.name}\" has been created" }
+    redirect_to(
+      edit_user_path(@guider),
+      success: "Slot Range \"#{SlotRangePresenter.new(@slot_range)}\" has been created"
```
benlovell: Same same.

app/controllers/users_controller.rb
```diff
@@ -7,6 +7,7 @@ def index
 
   def edit
     @guider = User.find(params[:id])
+    @slot_ranges = SlotRangePresenter.wrap(@guider.slot_ranges)
```
benlovell: Ooh la la.

app/controllers/slot_ranges_controller.rb
```diff
@@ -18,13 +18,19 @@ def update
       @slot_range.slots.destroy_all
       @slot_range.update(slot_range_parameters)
     end
-    redirect_to edit_user_path(@guider), flash: { success: "Guider \"#{@guider.name}\" has been updated" }
+    redirect_to(
+      edit_user_path(@guider),
+      success: 'Slot Range has been updated'
```
benlovell: Range is capitalised here but not below.

PR #19, opened 2016-09-29
Order slot ranges by from date

AndrewVos:


app/models/user.rb
```diff
@@ -13,6 +13,10 @@ def self.guiders
     User.all.select(&:guider?)
   end
 
+  def slot_ranges_for_display
```
benlovell: This feels like another presentation concern.

PR #20, opened 2016-09-29
Delete groups

benlovell:
Gives the resource manager the ability to delete groups.


app/controllers/groups_controller.rb
```diff
@@ -6,4 +6,11 @@ class GroupsController < ApplicationController
   def index
     @groups = Group.all
   end
+
+  def destroy
+    @tag = Tag.find(params[:id])
```
benlovell: Derp.

PR #26, opened 2016-09-29
Ensure Schedule#from is six weeks in the future

AndrewVos:
I am in the process of converting the schedules spec to use site_prism, but it's
becoming too much for one PR so I'll continue this afterwards.

### DIFF
--- a/app/views/schedules/_form.html.erb
+++ b/app/views/schedules/_form.html.erb
@@ -1,11 +1,22 @@
 <p>Click on the calendar to create available slots</p>
 
+<% if @schedule.errors.any? %>
+  <div class="alert alert-danger t-errors" role="alert">
+    <h3 class="alert__heading h4">There's a problem</h3>
+    <ul>
+      <% @schedule.errors.full_messages.each do |msg| %>
+        <li><%= msg %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+
 <%= form_for [@guider, @schedule], html: { class: "form-horizontal" } do |f| %>
   <div class="calendar">
     <div class="form-group">
-      <%= f.label :from, "From", class: 'col-sm-1 control-label' %>
+      <%= f.label :from, 'From', class: 'col-sm-1 control-label' %>
       <div class="col-sm-2">
-        <%= f.text_field :from, class: 'daterange form-control' %>
+        <%= f.text_field :from, class: 'daterange form-control t-from' %>
         <script>
         $(function() {
           $('.daterange').daterangepicker({
@@ -50,5 +61,5 @@
     </div>
   </div>
   <%= f.hidden_field :slots, class: 'slots-data', value: @schedule_json %>
-  <%= f.submit "Save", class: "btn btn-success" %>
+  <%= f.submit 'Save', class: 'btn btn-success t-save' %>
 <% end %>
--- a/spec/models/schedule_spec.rb
+++ b/spec/models/schedule_spec.rb
@@ -1,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe Schedule, type: :model do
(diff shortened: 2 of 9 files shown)


spec/support/pages/edit_schedule.rb
```diff
@@ -6,5 +6,10 @@ class EditSchedule < SitePrism::Page
       'h1',
       text: 'Sorry, you don\'t seem to have the resource_manager permission for this app.'
     )
+
+    elements :errors, '.field_with_errors'
+    element :error_summary, '.t-errors'
+
+    element :from, "input[name='schedule[from]']"
```
benlovell: Is there no way to tag this element with a `.t-*` style class?

spec/support/pages/new_schedule.rb
```diff
@@ -0,0 +1,11 @@
+module Pages
+  class NewSchedule < SitePrism::Page
+    set_url '/users/{user_id}/schedules/new'
+    element :save_button, "input[type='submit'][value=Save]"
```
benlovell: Same here and below?

spec/models/schedule_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Schedule, type: :model do
+  describe 'validation' do
```
benlovell: Should probably add an `valid with valid attributes` entry too. Just for a sanity check.

spec/models/schedule_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    context '#from' do
+      it 'is not nil' do
+        subject.from = nil
+        subject.valid?
+        expect(subject.errors[:from]).to_not be_empty
```
benlovell: Custom matcher for these might be nice instead of explicitly typing this stuff every time. Maybe something like `have_errors_on(:field_name)`

spec/models/schedule_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Schedule, type: :model do
+  describe 'validation' do
+    context '#from' do
+      it 'is not nil' do
+        subject.from = nil
+        subject.valid?
```
benlovell: It's exactly the same behaviour, but maybe more semantically correct to trigger this with `subject.validate` instead of the predicate.

spec/models/schedule_spec.rb
AndrewVos: done

spec/models/schedule_spec.rb
AndrewVos: done

spec/support/pages/new_schedule.rb
AndrewVos: done

spec/support/pages/edit_schedule.rb
AndrewVos: done

app/views/schedules/_form.html.erb
```diff
@@ -50,5 +61,5 @@
     </div>
   </div>
   <%= f.hidden_field :slots, class: 'slots-data', value: @schedule_json %>
-  <%= f.submit "Save", class: "btn btn-success" %>
+  <%= f.submit "Save", class: "btn btn-success t-save" %>
```
benlovell: Single quotes. Sorry. 😄 

Can we do them everywhere in this file?

spec/models/schedule_spec.rb
benlovell: Ah much nicer. 👍

app/views/schedules/_form.html.erb
AndrewVos: done

PR #30, opened 2016-09-30
Remove duplicate user authentication helper

benlovell:
We can use the one specified and mixed-in from the helper file.


Guntrisoft:
👍

PR #32, opened 2016-09-30
Fixing bug with searching for a guider's name

Guntrisoft:
- the markup for the glyphicon added was searchable aswell as the name


benlovell:
Whoops. I'm working on feature specs around this for guiders / groups so this won't happen again after that. 👍

PR #34, opened 2016-09-30
Install `phantomjs` this way instead

benlovell:
The other gem uses bitbucket for its CDN and hits the rate limits pretty
quickly and intermittently which is really nice when we're trying to get
our builds to run in CI.


Guntrisoft:
👍

PR #35, opened 2016-09-30
Refactor schedule and roles specs

AndrewVos:
~~Still adding site_prism so don't merge yet.~~

```
Resource manager manages groups
  Viewing existing groups
  Deleting a group

Resource manager manages schedules
  Successfully adds a new schedule
  Successfully updates a schedule
  Fails to create a schedule with invalid from date
  Fails to update a schedule with invalid from date

Roles
  Resource Managers
    Can manage guiders
    Can manage guiders schedules
    Can manage guiders slots
  Non-resource-managers
    Fail to manage guiders
    Fail to manage guiders schedules
    Fail to manage guiders slots

Schedule
  validation
    #from
      is valid with valid attributes
      is not nil
      is a minimum of six weeks in the future

User
  behaves like a gds-sso user class
    implements #where
    implements #update_attribute
    implements #update_attributes
    implements #create!
    implements #remotely_signed_out?
    the User class and GDS::SSO::User mixin work together
  #guider?
    normal user
      is false
    user with guider permission
      is true
  #resource_manager?
    normal user
      is false
    user with resource_manager permission
      is true
```

### DIFF
--- a/spec/features/resource_manager_manages_schedules_spec.rb
+++ b/spec/features/resource_manager_manages_schedules_spec.rb
@@ -7,9 +7,6 @@ RSpec.feature 'Resource manager manages schedules' do
   scenario 'Successfully adds a new schedule', js: true do
     given_the_user_is_a_resource_manager do
       and_there_is_a_guider
-      when_they_manage_guiders
-      then_they_see_the_guider_name_listed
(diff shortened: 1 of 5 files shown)


benlovell:
💯

PR #37, opened 2016-09-30
Use a human readable date format for Schedule#from

AndrewVos:


benbarnett:
We don't have to do this but GOVUK has a [suggested format](https://www.gov.uk/service-manual/user-centred-design/resources/patterns/dates.html#formatting-dates).

AndrewVos:
@benbarnett Thanks, added that

benbarnett:
Noice 👍

PR #38, opened 2016-09-30
Search and display a guider's groups

benlovell:
Displays group membership alongside the guider details. Also adds
searchability to the groups fields for guiders.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,6 +6,8 @@ class User < ApplicationRecord
 
   has_many :schedules, dependent: :destroy
 
+  has_and_belongs_to_many :groups # rubocop:disable Rails/HasAndBelongsToMany
+
   def self.guiders
     # This can't really be made faster because we're storing
     # permissions as a serialized string. We only really should have a maximum of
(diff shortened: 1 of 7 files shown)


benbarnett:
@benlovell the `.on(event, callback)` thing in the [list.js docs](http://www.listjs.com/docs/list-api) might be a useful hook for the test stuff?

app/models/user.rb
```diff
@@ -6,6 +6,8 @@ class User < ApplicationRecord
 
   has_many :schedules, dependent: :destroy
 
+  has_and_belongs_to_many :groups # rubocop:disable Rails/HasAndBelongsToMany
```
Guntrisoft: So whats the lowdown on why this needs to be disabled?

spec/spec_helper.rb
```diff
@@ -1,4 +1,6 @@
 RSpec.configure do |config|
+  config.order = :random
```
AndrewVos: Yes!

app/models/user.rb
benlovell: Rubocop wants you to use a `has_many :through` instead but this only makes sense when you want / care about attributes on the associative model, and we don't need those.

PR #39, opened 2016-09-30
First schedule can have any #from date

AndrewVos:
~~This is WIP at the moment because I'm not sure if the UI will work properly with
the code changes.~~

Feel free to review though!

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -5,9 +5,15 @@ class Schedule < ApplicationRecord
   scope :by_from, -> { order(:from) }
 
   validates :from, presence: true
-  validate :from_must_be_more_than_six_weeks_in_the_future
+  validate :from_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   private
+
+  def first_schedule?
+    user
+      .schedules
+      .none?(&:persisted?)
+  end
 
   def from_must_be_more_than_six_weeks_in_the_future
     maximum_age = 6.weeks.from_now.beginning_of_day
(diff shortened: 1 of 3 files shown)


app/models/schedule.rb
```diff
@@ -5,10 +5,17 @@ class Schedule < ApplicationRecord
...
   validates :from, presence: true
-  validate :from_must_be_more_than_six_weeks_in_the_future
+  validate :from_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   private
 
+  def first_schedule?
+    user
+      .schedules
+      .select(&:persisted?)
```
benlovell: You could replace the `select` `empty?` with `none?(&:persisted?)`.

spec/models/schedule_spec.rb
```diff
@@ -3,17 +3,48 @@
 RSpec.describe Schedule, type: :model do
   describe 'validation' do
     context '#from' do
-      it 'is valid with valid attributes' do
-        expect(build(:schedule, from: 6.weeks.from_now.beginning_of_day)).to be_valid
-      end
+      context 'first schedule' do
+        let :user do
```
benlovell: Nit-pick but can we wrap the `let` arguments in parens? This seems to be the prevalant style.

app/models/schedule.rb
AndrewVos: :+1:

spec/models/schedule_spec.rb
AndrewVos: done

PR #41, opened 2016-10-03
Resource managers can delete schedules

AndrewVos:
- Schedules can be deleted directly from the schedule list
- Schedules can NOT be deleted if they start from LESS than six weeks in
the future

image

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -30,11 +30,18 @@
       <td><span class="glyphicon glyphicon-calendar" aria-hidden="true"></span> <%= "#{schedule.title}" %></td>
       <td nowrap="true">
         <%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %>
-        <% if schedule.destroyable? %>
-          <%= link_to user_schedule_path(@guider, schedule), method: :delete, title: 'Delete Schedule', class: 'btn btn-danger t-delete' do %>
+        <div class="inline-button-to-container">
+          <%= button_to(
+              user_schedule_path(@guider, schedule),
+              method: :delete,
+              title: 'Delete Schedule',
+              class: 'btn btn-danger t-delete',
+              disabled: !schedule.destroyable?,
+              data: {confirm: 'Are you sure you want to delete this schedule?'}
+            ) do %>
             <span class="glyphicon glyphicon-remove" aria-hidden="true"></span><span class="sr-only">Delete Schedule</span>
           <% end %>
-        <% end %>
+        </div>
       </td>
     </tr>
   <% end %>
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -37,7 +37,7 @@
               title: 'Delete Schedule',
               class: 'btn btn-danger t-delete',
               disabled: !schedule.destroyable?,
-              data: {confirm: 'Are you sure you want to delete this schedule?'}
+              data: { confirm: "Are you sure you want to delete the schedule \"#{schedule.title}\"?" }
             ) do %>
             <span class="glyphicon glyphicon-remove" aria-hidden="true"></span><span class="sr-only">Delete Schedule</span>
           <% end %>
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -28,7 +28,7 @@
   <% @schedules.each do |schedule| %>
     <tr>
       <td><span class="glyphicon glyphicon-calendar" aria-hidden="true"></span> <%= "#{schedule.title}" %></td>
-      <td>
+      <td nowrap="true">
         <%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %>
         <% if schedule.destroyable? %>
           <%= link_to user_schedule_path(@guider, schedule), method: :delete, title: 'Delete Schedule', class: 'btn btn-danger t-delete' do %>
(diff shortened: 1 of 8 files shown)


app/views/users/edit.html.erb
```diff
@@ -15,8 +15,8 @@
 <table class="table table-striped table-bordered table-hover">
   <caption><span class="sr-only">List of guiders</span></caption>
   <colgroup>
-    <col width="98%">
-    <col width="1%">
+    <col width="91%">
+    <col width="9%">
```
benbarnett: If the screenshot is up to date, it might be worth giving the 'actions' column a bit more room so the edit & delete buttons can sit on the same line?

Guntrisoft: I've added a `nowrap="true"` to fix this

app/views/users/edit.html.erb
```diff
@@ -28,7 +28,18 @@
   <% @schedules.each do |schedule| %>
     <tr>
       <td><span class="glyphicon glyphicon-calendar" aria-hidden="true"></span> <%= "#{schedule.title}" %></td>
-      <td><%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %></td>
+      <td nowrap="true">
+        <%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %>
+        <% if schedule.destroyable? %>
+          <%= link_to user_schedule_path(@guider, schedule), method: :delete, title: 'Delete Schedule', class: 'btn btn-danger t-delete' do %>
```
benlovell: We should put a confirmation on this.

app/views/users/edit.html.erb
```diff
@@ -28,7 +28,18 @@
...
+            <span class="glyphicon glyphicon-remove" aria-hidden="true"></span><span class="sr-only">Delete Schedule</span>
+          <% end %>
+        <% else %>
+          <a href='javascript:void(0)' disabled title='Delete Schedule', class='btn btn-danger'>
```
benlovell: Double quote these attributes please. Also, is `javascript:void(0)` something we do? Could we just leave the `href` off or maybe even just link it to a fragment `#` or something.

spec/features/resource_manager_manages_schedules_spec.rb
```diff
@@ -95,11 +121,14 @@ def then_they_are_told_that_the_schedule_has_been_created
     @page = Pages::EditUser.new
     expect(@page).to have_flash_of_success
   end
-
-  def then_they_are_told_that_the_schedule_has_been_updated
-    @page = Pages::EditUser.new
-    expect(@page).to have_flash_of_success
-  end
+  alias_method(
```
benlovell: Maybe just drop the parentheses and have the first argument on the same line as `alias_method` and each successive argument on a new line? Seems a little neater than this style to my eyes. But i'm not overly fussed on this.

spec/support/pages/edit_user.rb
```diff
@@ -7,5 +7,6 @@ class EditUser < SitePrism::Page
       text: 'Sorry, you don\'t seem to have the resource_manager permission for this app.'
     )
     element :flash_of_success, '.alert-success'
+    element :delete_button, '.t-delete'
```
benlovell: I think we can drop the name of the thing. Gives us scope to change the thing if we need to and usually the naming is enough context.

app/views/users/edit.html.erb
benbarnett: Ideally both delete buttons should be a `<button>`, as semantically we're performing an action rather than navigating to a new page. Browsers also honour the `disabled` attribute on those elements so we can avoid the href weirdness.

spec/features/resource_manager_manages_schedules_spec.rb
AndrewVos: Rubocop doesn't like that actually. Arguments need to be aligned

app/views/users/edit.html.erb
Guntrisoft: agreed

app/views/users/edit.html.erb
AndrewVos: Done

spec/support/pages/edit_user.rb
AndrewVos: done

app/views/users/edit.html.erb
```diff
@@ -28,7 +28,24 @@
...
-      <td><%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %></td>
+      <td nowrap="true">
+        <%= link_to edit_user_schedule_path(@guider, schedule), title: "Edit Schedule #{schedule.title}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-pencil" aria-hidden="true"></span><span class="sr-only">Edit Schedule</span><% end %>
+        <% if schedule.destroyable? %>
+          <%= link_to(
+              user_schedule_path(@guider, schedule),
+              method: :delete,
+              title: 'Delete Schedule',
+              class: 'btn btn-danger t-delete',
+              data: {confirm: 'Are you sure you want to delete this schedule?'}
```
benlovell: The house-style is to include spaces in the hash braces. Also, we should confirm with the title of the schedule so they know if they've clicked the wrong one.

spec/features/resource_manager_manages_schedules_spec.rb
benlovell: Ugh. OK.

app/views/users/edit.html.erb
AndrewVos: Good idea, done.

app/views/users/edit.html.erb
AndrewVos: Good idea thanks! Done.

PR #44, opened 2016-10-03
Show Schedule end date (one day before next schedule start)

AndrewVos:
To anyone looking at this, the important code is in the first commit.

image

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -2,22 +2,22 @@ class Schedule < ApplicationRecord
   belongs_to :user
   has_many :slots, inverse_of: :schedule, dependent: :destroy
   accepts_nested_attributes_for :slots, allow_destroy: true
-  scope :by_from, -> { order(:from) }
+  scope :by_start_at, -> { order(:start_at) }
 
-  validates :from, presence: true
-  validate :from_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
+  validates :start_at, presence: true
+  validate :start_at_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   def end_at
     self[:end_at] - 1.day if self[:end_at]
   end
 
   def self.with_end_at
-    select('*, LEAD("from", 1) OVER (ORDER BY "from") AS end_at')
-      .by_from
+    select('*, LEAD(start_at, 1) OVER (ORDER BY start_at) AS end_at')
+      .by_start_at
   end
 
   def modifiable?
-    from > 6.weeks.from_now
+    start_at > 6.weeks.from_now
   end
 
   private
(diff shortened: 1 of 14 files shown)


benbarnett:
🤘

app/models/schedule.rb
```diff
@@ -2,13 +2,22 @@ class Schedule < ApplicationRecord
...
-  scope :by_from, -> { order(:from) }
+  scope :by_start_at, -> { order(:start_at) }
 
-  validates :from, presence: true
-  validate :from_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
+  validates :start_at, presence: true
+  validate :start_at_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
+
+  def end_at
+    self[:end_at] - 1.day if self[:end_at]
```
benlovell: Can we `super - 1.day if super` instead here?

app/models/schedule.rb
```diff
@@ -2,13 +2,22 @@ class Schedule < ApplicationRecord
...
+  end
+
+  def self.with_end_at
+    select('*, LEAD(start_at, 1) OVER (ORDER BY start_at) AS end_at')
+      .order(:start_at)
```
benlovell: Can you chain scopes off `select` I can't remember? If so, we could use the `by_start_at` scope.

spec/models/schedule_spec.rb
```diff
@@ -2,17 +2,17 @@
...
+    context '#start_at' do
       context 'first schedule' do
         let(:user) do
           create(:user)
         end
 
-        def build_schedule(from)
-          build(:schedule, user: user, from: from)
+        def build_schedule(start_at)
+          build(:schedule, user: user, start_at: start_at)
```
benlovell: General observation: for lots of these you could use `build_stubbed` instead as they don't roundtrip.

app/models/schedule.rb
dwhenry: `LEAD(start_at, 1)` == `LEAD(start_at)` also what happens if this is called without the schedule being scoped to a guider.. can this happen?

AndrewVos: @dwhenry That's not something that happens yet (or ever, possibly).

app/models/schedule.rb
AndrewVos: We can't actually, because of the way the code is now. There is no :end_at accessor defined. I've tried defining one and it's always nil even if self[:end_at] is not. Bit confused about how rails sets those ivars.

spec/models/schedule_spec.rb
AndrewVos: Yeah good point. Totally forgot about that.

app/models/schedule.rb
benlovell: Oh wurd.

app/models/schedule.rb
AndrewVos: @dwhenry Oh I remove the 1 btw.. thanks :)

AndrewVos: @benlovell Also I chained that scope...

dwhenry: I know it's a little late to the party but changing it to the following should remove any future issues with guider scope:
```
LEAD(start_at) OVER (PARTITION BY guider_id ORDER BY start_at) AS end_at
```

PR #48, opened 2016-10-04
Appointments

AndrewVos:
This needs some heavy refactoring which I'm working on at the moment.

I would appreciate some brutal code review though.

(screenshots are zoomed out because it doesn't all fit on my tiny screen)

image

image


app/controllers/appointments_controller.rb
```diff
@@ -0,0 +1,53 @@
+class AppointmentsController < ApplicationController
+  before_action :authorise_for_contact_centre_agents!
+
+  def new
+    UsableSlot.regenerate_for_six_weeks
```
AndrewVos: This should be replaced with a background job.

app/controllers/schedules_controller.rb
```diff
@@ -52,9 +52,11 @@ def load_current_guider
   def schedule_json
```
benlovell: This doesn't feel like a controller concern now. Can we move it to a dedicated serialiser?

app/models/appointment.rb
```diff
@@ -0,0 +1,19 @@
+class Appointment < ApplicationRecord
+  belongs_to :user
```
benlovell: I wonder if it would be more helpful to refer to `guider` here instead? It seems we have a bunch of places now where we could provide these useful cues to disambiguate.

app/models/usable_slot.rb
```diff
@@ -0,0 +1,73 @@
+class UsableSlot < ApplicationRecord
+  belongs_to :user
```
benlovell: Same here with referring to `guider` instead.

app/models/usable_slot.rb
```diff
@@ -0,0 +1,73 @@
...
+      .select('DISTINCT usable_slots.start_at, usable_slots.end_at, count(1) AS guiders')
+      .without_appointments
+      .group('usable_slots.start_at, usable_slots.end_at')
+      .within_date_range(from, to)
+      .map do |us|
+      { guiders: us.attributes['guiders'], start: us.start_at, end: us.end_at }
+    end
+  end
+
+  private_class_method
```
benlovell: This has no effect unless you pass the method names as symbols. It's not actually a language keyword as `private` is, just a regular method.

db/migrate/20161004141342_break_apart_slots_start_and_end.rb
```diff
@@ -0,0 +1,45 @@
+class BreakApartSlotsStartAndEnd < ActiveRecord::Migration[5.0]
+  def up
+    add_column :slots, :day_of_week, :integer
+    add_column :slots, :start_hour, :integer
+    add_column :slots, :start_minute, :integer
+    add_column :slots, :end_hour, :integer
+    add_column :slots, :end_minute, :integer
+
+    Slot.all.each do |slot|
```
benlovell: Is this stuff necessary right now? We don't have any data that isn't throwaway.

db/migrate/20161004141342_break_apart_slots_start_and_end.rb
```diff
@@ -0,0 +1,45 @@
...
+    remove_column :slots, :start_at
+    remove_column :slots, :end_at
+  end
+
+  def down
+    add_column :slots, :day, :string
+    add_column :slots, :start_at, :string
+    add_column :slots, :end_at, :string
+
+    Slot.all.each do |slot|
```
benlovell: Same here of course.

db/migrate/20161005131704_create_usable_slots.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUsableSlots < ActiveRecord::Migration[5.0]
+  def change
+    create_table :usable_slots do |t|
```
benlovell: Can we add indexes for the foreign keys?

spec/features/contact_centre_agent_creates_appointments_spec.rb
```diff
@@ -0,0 +1,69 @@
+# rubocop:disable Metrics/MethodLength
+require 'rails_helper'
+
+RSpec.feature 'Contact centre agent creates appointments' do
```
benlovell: I wonder if we can shorten _contact centre agent_ to _agent_ without losing any meaning in the domain? cc @benbarnett 

If we agree on this we should shorten it everywhere it's currently referred to as _contact centre agent_.

spec/features/resource_manager_manages_schedules_spec.rb
```diff
@@ -156,14 +156,22 @@ def and_the_guider_has_the_changed_time_slots
     @schedule.reload
 
     first_slot = @schedule.slots.first
-    expect(first_slot.day).to eq 'Wednesday'
-    expect(first_slot.start_at).to eq '11:00'
-    expect(first_slot.end_at).to eq '12:10'
+    expect(first_slot.attributes).to include(
```
benlovell: You could tidy this up a little with the `have_attributes` matcher. At the very least it would allow you to use symbol keys.

spec/features/roles_spec.rb
```diff
@@ -26,23 +26,39 @@
     end
   end
 
-  context 'Non-resource-managers' do
-    scenario 'Fail to manage guiders' do
+  context 'Contact Centre Agents' do
+    scenario 'Can book appointments' do
+      given_the_user_is_a_contact_centre_agent do
```
benlovell: Wut?

spec/models/schedule_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe Schedule, type: :model do
+  include ActiveSupport::Testing::TimeHelpers
```
benlovell: May as well move this up into the `rails_helper`.

app/models/appointment.rb
AndrewVos: Yes that thought crossed my mind a lot. Is it worth it having to refer to the class name directly? I think so.

app/models/usable_slot.rb
AndrewVos: Yeah I actually just did what rubocop said there without even learning about `private_class_method`. I must google this

db/migrate/20161004141342_break_apart_slots_start_and_end.rb
AndrewVos: Hm, I guess not. I could just drop everything from the table to be sure.

app/models/appointment.rb
benlovell: Depending on how far we want to take it we could just derive a `Guider` from `User`. Then the guider-specific stuff could live there at least... Hmm.

spec/features/roles_spec.rb
AndrewVos: Weird, fixed!

app/controllers/schedules_controller.rb
AndrewVos: Done here: https://github.com/guidance-guarantee-programme/telephone_appointment_planner/pull/48/commits/ad53c5df1970dc81776831b7ce3f990a026d6e69

spec/features/contact_centre_agent_creates_appointments_spec.rb
benbarnett: Yep, that seems cool. Just had a chat with @mattlucht about it and it seems 'guiders', 'agents' and 'resource managers' are clear enough names.

spec/models/schedule_spec.rb
AndrewVos: Done

spec/features/resource_manager_manages_schedules_spec.rb
AndrewVos: Nice, forgot about that! Done.

db/migrate/20161004141342_break_apart_slots_start_and_end.rb
AndrewVos: done

db/migrate/20161005131704_create_usable_slots.rb
AndrewVos: done

PR #51, opened 2016-10-04
Create / assign group to guiders

benlovell:
Creates (if necessary) and assigns the group to the specified guiders.
This will always appear successful as when there is already an
associated group the guider is simply skipped over.


app/models/group.rb
```diff
@@ -1,3 +1,12 @@
 class Group < ApplicationRecord
   default_scope { order(:name) }
+
+  has_many :group_assignments, class_name: GroupAssignment
```
dwhenry: if you are using `class_name: ` it should be a string to avoid load order issues.

app/models/group_assignment.rb
```diff
@@ -0,0 +1,6 @@
+class GroupAssignment < ApplicationRecord
+  self.table_name = :groups_users
```
dwhenry: This seems like it will be really annoying later on. Any reason we have decided on this approach?

spec/forms/create_group_assignments_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  subject { described_class.new(user_ids, group_params) }
+
+  context 'when the users are not already members' do
+    it 'assigns the group to the given users' do
+      # ensure just the one group is created
+      expect { subject.call }.to change { Group.count }.by(1)
+
+      users.each do |user|
+        expect(user.groups.map(&:name)).to eq(['Another Group'])
+      end
```
dwhenry: is this two tests?

benlovell: You could definitely argue that. I collapsed them though as the one logical assertion per spec gets stupid-slow after a while 😢

app/models/group.rb
benlovell: Autoloading no longer happens in production after app boot (since Rails 5)

app/models/group_assignment.rb
benlovell: Actually on reflection I think this should remain. The association table is `groups_users` as those are the tables we're linking, but this doesn't make for a descriptive name that fits the domain language on the associative model. `GroupAssignment` accurately describes this I feel.

app/models/group.rb
dwhenry: is the `class_name: ` needed now the table name matches the class name?

benlovell: Derp. Missed that - resolved now.

dwhenry:
LGTM but would be good to get the JS/CSS review by someone how is closer to the project. :)

app/assets/javascripts/guiders.js
```diff
@@ -0,0 +1,42 @@
+$(document).on('turbolinks:load', function() {
+  $('.js-go-link').on('click', function() {
```
AndrewVos: You can use `$(document).on(".js-go-link", "click", fn)` sort of style instead of waiting for `turbolinks:load`. In my experience this produces much more robust javascript.

AndrewVos: If you used `$(document).on` you could move everything in this file up out of the `turbolinks:load`.

benlovell: @andrewvos cool, might be worth holding fire on the JS commentary as @guntrisoft is giving it a full going over tomorrow.

AndrewVos:
good to merge

benlovell:
Is there no end to this man's talent? @Guntrisoft

PR #52, opened 2016-10-05
Unassign guiders from groups

benlovell:
For the moment this leaves the actual group instances around.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -17,6 +17,15 @@ class GroupsController < ApplicationController
     )
   end
 
+  def destroy
+    DestroyGroupAssignments.new(user_ids, params[:id]).call
+
+    redirect_back(
+      fallback_location: users_path,
+      success: 'The users were removed from the specified groups'
+    )
+  end
+
   private
 
   def group_params
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -10,23 +10,23 @@ class GroupsController < ApplicationController
 
   def create
     CreateGroupAssignments.new(user_ids, group_params).call
-
-    redirect_back(
-      fallback_location: users_path,
-      success: 'The users were assigned to the specified groups'
-    )
+    go_back_with_success(:assigned)
   end
 
   def destroy
     DestroyGroupAssignments.new(user_ids, params[:id]).call
 
-    redirect_back(
-      fallback_location: users_path,
-      success: 'The users were removed from the specified groups'
-    )
+    go_back_with_success(:removed)
   end
 
   private
+
+  def go_back_with_success(action)
+    redirect_back(
+      fallback_location: users_path,
+      success: "The users were #{action} from the specified groups"
+    )
+  end
 
   def group_params
     params.require(:group).permit(:name)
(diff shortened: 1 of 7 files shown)


app/controllers/groups_controller.rb
```diff
@@ -10,15 +10,24 @@ def index
...
+
+  def destroy
+    DestroyGroupAssignments.new(user_ids, params[:id]).call
+
+    go_back_with_success(:removed)
+  end
 
+  private
+
+  def go_back_with_success(action = :assigned)
```
AndrewVos: I would make this parameter not optional so it's more obvious when you use it in `create`.

app/forms/destroy_group_assignments.rb
```diff
@@ -0,0 +1,15 @@
+class DestroyGroupAssignments
+  attr_reader :user_ids
+  attr_reader :group_id
```
AndrewVos: Not sure you need to expose these

spec/features/resource_manager_manages_guiders_spec.rb
```diff
@@ -23,6 +23,30 @@
...
+    and_they_choose_to_add_groups_to_multiple_guiders
+  end
+
+  def when_they_remove_a_group
+    @page.groups.first.remove.click
+  end
+
+  def then_the_group_is_unassigned_from_those_guiders
+    expect(@page).to have_flash_of_success
+    expect(@page).to have_no_text(@team_one.name)
```
AndrewVos: Tempting to check the groups are removed in the database too

app/controllers/groups_controller.rb
benlovell: Doneski.

spec/features/resource_manager_manages_guiders_spec.rb
AndrewVos: Oh I read the description now. Why is that?

benlovell: They could be reassigned in future as providing the name of an existing group would use that particular instance. Also it cuts out on the need for weird logic checking that the group doesn't have associated assignments and deleting on that basis.

I'm adding a typeahead on the group text field so this would expose the existing groups too.

AndrewVos:
Cool cool merge it

app/forms/destroy_group_assignments.rb
```diff
@@ -0,0 +1,17 @@
...
+  def call
+    GroupAssignment
+      .where(user_id: user_ids, group_id: group_id)
+      .delete_all
+  end
+
+  private
+
+  attr_reader :user_ids
+  attr_reader :group_id
```
dwhenry: are these needed? 
cleaner to just used @user_ids and @group_id

app/controllers/groups_controller.rb
```diff
@@ -10,15 +10,24 @@ def index
 
   def create
     CreateGroupAssignments.new(user_ids, group_params).call
+    go_back_with_success(:assigned)
```
dwhenry: ```"The users were assigned from the specified groups"```
should this be changed to:
```go_back_with_success('assigned to')```

PR #57, opened 2016-10-07
Toggle javascript 'ready' event based on Turbolinks support

benbarnett:
The HTML5 History API is required for Turbolinks. This change checks if Turbolinks is being run in a supported browser and toggles the event we wait to boot the javascript with accordingly.


Guntrisoft:
👍

PR #60, opened 2016-10-10
Creating a graphical view of schedules

Guntrisoft:
screen shot 2016-10-10 at 11 02 39

- shows six week window and visual indicates whether a schedule is editable or not
- altered schedule model's end_at method to take only a second off rather than a day
- improved wording of uneditable schedule sentence

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -8,7 +8,7 @@ class Schedule < ApplicationRecord
   validate :start_at_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   def end_at
-    self[:end_at] - 1.day if self[:end_at]
+    self[:end_at] - 1.second if self[:end_at]
   end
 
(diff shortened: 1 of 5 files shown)


app/models/schedule.rb
```diff
@@ -8,7 +8,7 @@ class Schedule < ApplicationRecord
   validate :start_at_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   def end_at
-    self[:end_at] - 1.day if self[:end_at]
+    self[:end_at] - 1.second if self[:end_at]
```
Guntrisoft: Would be good to get your opinion on this @AndrewVos - seems to work for what I want - just want to be sure it won't affect anything else.

AndrewVos: I think this should work :)

AndrewVos:
lgtm, very cool!

PR #61, opened 2016-10-10
Schedule default start is (six weeks + one day)

AndrewVos:
This is because when we create a new schedule, we still want to be able
to edit that schedule today.

https://trello.com/c/GRqNIx44/116-guiders-cannot-have-more-than-one-schedule-starting-on-the-same-date

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -3,9 +3,7 @@ class SchedulesController < ApplicationController
   before_action :load_current_guider
 
   def new
-    @schedule = @guider.schedules.build(
-      start_at: 6.weeks.from_now
-    )
+    @schedule = @guider.schedules.build
     @slots_as_json = slots_as_json
   end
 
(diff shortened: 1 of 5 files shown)


app/controllers/schedules_controller.rb
```diff
@@ -4,7 +4,7 @@ class SchedulesController < ApplicationController
 
   def new
     @schedule = @guider.schedules.build(
-      start_at: 6.weeks.from_now
+      start_at: 6.weeks.from_now + 1.day
```
benlovell: I think we should move this logic into `User` or perhaps even within the scope of `User#schedules` (you can pass a block of methods that are executed in that scope.) 

It seems like a domain thing that should be enforced and is somewhat lost out here in this controller.

AndrewVos: 100% agree. Was thinking something like:

```ruby
class Schedule
  def self.build_with_future_start_date
```

benlovell: Makes sense.

AndrewVos: Ok I couldn't do that without patching Relation. See the latest force push though. @benlovell

PR #65, opened 2016-10-10
Rename usable_slots to bookable_slots

AndrewVos:


app/models/bookable_slot.rb
```diff
@@ -8,26 +8,26 @@ def self.regenerate_for_six_weeks
   end
 
   def self.within_date_range(from, to)
-    where('usable_slots.start_at > ? AND usable_slots.end_at < ?', from, to)
+    where('bookable_slots.start_at > ? AND bookable_slots.end_at < ?', from, to)
```
benlovell: In lots of cases where we're explicitly providing the table we don't actually need to, thus it makes changes like this more work than renaming the table / class.

AndrewVos: Problem is, when combining this that `LEFT JOIN` further up we'll have problems. Feels safe to be explicit, or at least interpolate the table name in there.

benlovell: Ah, yeah if they're chained that is an issue. Perhaps we should use `quoted_table_name` in place of the explicit table naming then... I'm merging this for now anyway.

PR #66, opened 2016-10-10
Improving layout of new appointment page

Guntrisoft:
screen shot 2016-10-10 at 15 14 15

- making calendar go full width
- personal details only in two columns
- improved vertical spacing between sections of form


benlovell:
Typo: improving

benlovell:
😍

PR #69, opened 2016-10-10
Increase size of notes field on appointment/new

Guntrisoft:
- Move mobile to first column to balance columns


Guntrisoft:
@benbarnett

benbarnett:
👍

PR #75, opened 2016-10-11
Colour coded availability

Guntrisoft:
screen shot 2016-10-11 at 11 39 40

### DIFF
--- a/app/assets/javascripts/calendars/appointment-availability.es6
+++ b/app/assets/javascripts/calendars/appointment-availability.es6
@@ -28,26 +28,26 @@
             'max-width': '46%',
             'padding': '.2em',
             'box-sizing': 'border-box',
-            'cursor': 'pointer'
+            'cursor': 'pointer',
+            'background': this.getEventColour(event)
           });
         },
         eventClick: (event) => {
-          const events = this.$el.fullCalendar('clientEvents'),
-                currentBackgroundColor = event.backgroundColor;
+          const events = this.$el.fullCalendar('clientEvents');
 
           let start = '',
                 end = '';
 
           for (let eventIndex in events) {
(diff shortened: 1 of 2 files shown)


AndrewVos:
Wooo I like it

app/assets/javascripts/calendars/appointment-availability.es6
```diff
@@ -59,19 +59,46 @@
...
       this.init();
       this.addEvents();
     }
 
+    getEventColour(event) {
+      if (event.selected === true) {
+        return this.selectedEventColour;
+      }
+
+      const colours = $.grep(this.coloursConfig, (configItem) => {
```
dwhenry: what happens if an event has more than 30 guiders?

could this be written using `find` instead of `grep` as that feels like it should be easier?

Guntrisoft: Think we'll have to play it by ear with the total - at the moment they have an overlap of 15 guiders in the early shift with 15 in the mid shift that do a 12.20 appointment, making up to 30 guiders at one time available - if they do have more guiders available at one particular time it just means that slot will be dark blue.

`find` is used to filter DOM elements, rather than `$.grep()` which is used to filter an array.

benbarnett: Yeah I don't think we need to worry too much about the resolution of the colour bands. If there's a slight precedence to book one with more availability then great, but the customer is still able to book any slot that's available regardless.

Guntrisoft: Nope you're not missing something, you're right about that! - Created https://github.com/guidance-guarantee-programme/telephone_appointment_planner/pull/80 to fix it

PR #77, opened 2016-10-11
Force a sort on initialisation of sortable list

Guntrisoft:
- this fixes an issue when you click the 'name' header for
  the first time and it doesn't sort as the list is already
  sorted A-Z in the DOM when it's initialised
- fixed a bug in the config for the sortable list
  when there is no groups (trailing comma)

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -1,7 +1,7 @@
 <h1>Manage Guiders</h1>
 
 <div data-module="GuidersMultiAction">
-  <div id="guiders" data-module="SortableList" data-config='{"valueNames": ["name", <%= @groups.map { |group| "\"#{dom_id(group)}\"" }.join(', ') %>]}'>
+  <div id="guiders" data-module="SortableList" data-default-order='{"value": "name", "order": "asc"}' data-config='<%= sortable_list_config(@groups) %>'>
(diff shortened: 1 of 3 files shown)


app/views/users/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <h1>Manage Guiders</h1>
 
 <div data-module="GuidersMultiAction">
-  <div id="guiders" data-module="SortableList" data-config='{"valueNames": ["name", <%= @groups.map { |group| "\"#{dom_id(group)}\"" }.join(', ') %>]}'>
+  <div id="guiders" data-module="SortableList" data-default-order='{"value": "name", "order": "asc"}' data-config='{"valueNames": ["name"<%= @groups.any? ? ',':'' %><%= @groups.map { |group| "\"#{dom_id(group)}\"" }.join(', ') %>]}'>
```
Guntrisoft: @benlovell anything nicer than this I can do - it looks a mess in the markup!

benlovell: I'd say it warrants moving to a helper function now instead.

Guntrisoft: Done

Guntrisoft:
@benlovell happy with the view helper?

PR #78, opened 2016-10-12
Add 'Who is your pension provider?' field to appointments/new

AndrewVos:


app/helpers/appointment_helper.rb
```diff
@@ -17,13 +17,52 @@ module AppointmentHelper
...
+      WHO_IS_YOUR_PENSION_PROVIDER_OPTIONS,
+      appointment
+    )
+  end
+
+  private
+
+  def select_options(field, options, object)
+    options |=
+      [object.send(field)]
```
benlovell: We should favour `public_send` here and below.

AndrewVos: Done, and also reordered the method parameters to be `(object, field, options)`. Seemed more logical.

dwhenry: Do we need to use send. 

Would it be easier to just pass the value in given we only use the field to extract the value from the object (which we also pass in)

AndrewVos: Yeah that sounds good to me... Done. @dwhenry

PR #79, opened 2016-10-12
Replace npm with yarn

AndrewVos:
Could one or two of you test this on your machine before we start using it in prod


.travis.yml
```diff
@@ -6,12 +6,17 @@ cache:
   bundler: true
   directories:
   - tmp/cache/assets/test
+  - $HOME/.nvm
+  - $HOME/.yarn-cache
 node_js:
```
Guntrisoft: Is this actually used, seeing as it's a ruby project?

benbarnett: Nopers, we had it in the main frontend and it was just tricking us the whole time

benbarnett: https://github.com/guidance-guarantee-programme/pension_guidance/commit/ad872e9598b0d5c3885b0883fdb3d0d44cf95399

PR #81, opened 2016-10-12
[WIP DO NOT MERGE] - Guiders individual calendar and company wide calendar

Guntrisoft:
- Using fullcalendar-scheduler
- Using fullcalendar on asset-rails instead as
  scheduler is available there too as a rails gem


Guntrisoft:
Deleting this as it's not needed anymore

PR #82, opened 2016-10-12
Change appointments year_of_birth to date_of_birth

AndrewVos:

### DIFF
--- a/db/migrate/20161012143328_add_date_of_birth_to_appointments.rb
+++ b/db/migrate/20161012143328_add_date_of_birth_to_appointments.rb
@@ -1,0 +1,5 @@
+class AddDateOfBirthToAppointments < ActiveRecord::Migration[5.0]
+  def change
+    add_column :appointments, :date_of_birth, :date
+  end
+end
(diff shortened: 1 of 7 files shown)


benbarnett:
How does this render? Three dropdowns?

AndrewVos:
@benbarnett Kind of. It's just the html5 picker. Attached screenshot ^^

db/migrate/20161012143328_add_date_of_birth_to_appointments.rb
```diff
@@ -0,0 +1,5 @@
+class AddDateOfBirthToAppointments < ActiveRecord::Migration[5.0]
+  def change
+    add_column :appointments, :date_of_birth, :datetime
```
dwhenry: why are we storing DOB as a datetime?

AndrewVos: Wow good catch! Sorted.

spec/features/contact_centre_agent_creates_appointments_spec.rb
```diff
@@ -61,7 +61,7 @@ def then_that_appointment_is_stored
     expect(appointment.email).to eq 'email@example.org'
     expect(appointment.phone).to eq '0000000'
     expect(appointment.mobile).to eq '1111111'
-    expect(appointment.year_of_birth).to eq '1891'
+    expect(appointment.date_of_birth).to eq '1983-10-23 00:00:00.000000000 +0000'
```
Guntrisoft: Should this still be in this format now it's a date?

AndrewVos: Yup yup was just fixing that :)

PR #84, opened 2016-10-13
Convert select2 dropdown to standard module

Guntrisoft:


app/views/appointments/new.html.erb
```diff
@@ -39,7 +39,7 @@
             label: 'Where did you hear about Pension Wise?',
             include_blank: true
           },
-          { class: 'select2 t-where-did-you-hear-about-pension-wise' }
+          { class: 't-where-did-you-hear-about-pension-wise', 'data-module': 'AdvancedSelect', 'data-config': '{"tags": true, "placeholder": "Type a value in here if it is not included in the list"}' }
```
AndrewVos: You don't have to put that data into a string, you can use a ruby hash there:

```erb
{ class: 't-where-did-you-hear-about-pension-wise', data: {module: 'AdvancedSelect', config: {tags: true, placeholder: "Type a value in here if it is not included in the list"} }
```

PR #85, opened 2016-10-13
Fix pension provider dropdown to work as a select2 dropdown

Guntrisoft:
- moved Year of birth to left column to even out form


Guntrisoft:
@AndrewVos

PR #86, opened 2016-10-13
Appointments cannot be booked < 2 working days, > 30 working days

AndrewVos:
Not taking into account bank holidays at the moment, which should be
worked on when we do the non-working-days feature.

This should abide by the rules mentioned here:
https://trello.com/c/XZoLbCJZ/119-appointment-booking-part-2

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -9,6 +9,9 @@ class Appointment < ApplicationRecord
   validates :memorable_word, presence: true
   validates :where_did_you_hear_about_pension_wise, presence: true
 
+  validate :not_within_two_business_days
+  validate :not_more_than_thirty_business_days_in_future
+
   def assign_to_guider
     slot = BookableSlot
            .without_appointments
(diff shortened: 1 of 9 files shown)


app/models/appointment.rb
```diff
@@ -17,4 +20,18 @@ def assign_to_guider
            .first
     self.guider = slot.guider if slot
   end
+
+  private
+
+  def not_within_two_business_days
+    return unless start_at.present?
```
benlovell: I think just `return unless start_at` would suffice here and below. Also can you put a line of whitespace as breathing room between these guard clauses and the rest of the body?

spec/features/contact_centre_agent_creates_appointments_spec.rb
```diff
@@ -2,6 +2,10 @@
 require 'rails_helper'
 
 RSpec.feature 'Agent creates appointments' do
+  let :day do
```
benlovell: We generally use parens around the arguments to `let`... Also don't be afraid of the single line and braces style if the thing is short enough. 😉

spec/models/appointment_spec.rb
```diff
@@ -2,9 +2,12 @@
 
 RSpec.describe Appointment, type: :model do
   describe 'validations' do
+    let(:subject) do
+      build(:appointment)
```
benlovell: This could be `build_stubbed` I think.

config/business_time.yml
```diff
@@ -0,0 +1,10 @@
+business_time:
+  beginning_of_workday: 9:00 am
```
AndrewVos: It's not important at the moment, but it might become important. Updated.

spec/models/appointment_spec.rb
AndrewVos: Done

spec/features/contact_centre_agent_creates_appointments_spec.rb
AndrewVos: Hard to remember that one! done

app/models/appointment.rb
AndrewVos: Done

AndrewVos:
@benlovell Addressed all your comments ^

PR #89, opened 2016-10-14
Customers can attempt an appointment and either be eligible or not

AndrewVos:
- Seperate page to take first_name, last_name, date_of_birth, and
defined_contribution_pension.
- If they are not eligible the agent is told.
- If they are then they can continue to book an appointment as usual.

(note that those fields are prefilled)

### DIFF
--- a/app/controllers/appointment_attempts_controller.rb
+++ b/app/controllers/appointment_attempts_controller.rb
@@ -1,0 +1,40 @@
+class AppointmentAttemptsController < ApplicationController
+  before_action :authorise_for_agents!
+
+  def new
+    @appointment_attempt = AppointmentAttempt.new
+  end
+
+  def create
+    @appointment_attempt = AppointmentAttempt.new(appointment_attempt_params)
+    if @appointment_attempt.save
+      if @appointment_attempt.eligible?
+        redirect_to new_appointment_attempt_appointment_path(@appointment_attempt)
+      else
+        redirect_to ineligible_appointment_attempts_path
+      end
+    else
+      render :new
+    end
+  end
+
+  def ineligible
+  end
+
+  private
+
+  def appointment_attempt_params
+    params
+      .require(:appointment_attempt)
+      .permit(
+        :first_name,
+        :last_name,
+        :date_of_birth,
+        :defined_contribution_pot
+      )
+  end
+
+  def authorise_for_agents!
+    authorise_user!(User::AGENT_PERMISSION)
+  end
+end
--- a/app/models/appointment_attempt.rb
+++ b/app/models/appointment_attempt.rb
@@ -1,0 +1,9 @@
+class AppointmentAttempt < ApplicationRecord
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :date_of_birth, presence: true
+
+  def eligible?
+    defined_contribution_pot? && ((date_of_birth + 50.years) < Time.zone.today)
+  end
+end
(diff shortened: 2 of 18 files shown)


app/controllers/appointment_attempts_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      else
+        redirect_to not_eligible_path
+      end
+    else
+      render :new
+    end
+  end
+
+  def not_eligible
+    render 'not_eligible'
```
benlovell: This line isn't necessary. This will happen implicitly.

app/views/appointment_attempts/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Book an appointment</h1>
+
+<%= form_for @appointment_attempt, layout: :basic do |f| %>
+  <%= f.text_field :first_name, class: 't-first-name' %>
+  <%= f.text_field :last_name, class: 't-last-name' %>
+  <%= f.date_field :date_of_birth, class: 't-date-of-birth', data: { module: 'CustomerAge', output: '.customer-age' } %>
+  <div class="bg-info customer-age">
```
benbarnett: Tiny thing but we try to prefix javascript targets with 'js-'  which then aren't used for styling. Just means we can refractor css without breaking js.

app/models/appointment_attempt.rb
```diff
@@ -0,0 +1,9 @@
+class AppointmentAttempt < ApplicationRecord
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :date_of_birth, presence: true
+
+  def eligible?
+    defined_contribution_pot && ((date_of_birth + 50.years) < Time.zone.today)
```
benlovell: Can we use the predicate: `defined_contribution_pot?` instead?

app/views/layouts/application.html.erb
```diff
@@ -9,7 +9,7 @@
     <%= active_link_to 'Manage guiders', users_path, wrap_tag: :li %>
   <% end %>
   <% if current_user.agent? %>
-    <%= active_link_to 'Create appointments', new_appointment_path, wrap_tag: :li %>
+    <%= active_link_to 'Create appointments', new_appointment_attempt_path, wrap_tag: :li %>
```
benbarnett: Should probably just have the singular for this?

config/routes.rb
```diff
@@ -10,7 +10,10 @@
   end
 
   resources :groups, only: %i(index create destroy)
-  resources :appointments
+  resources :appointment_attempts do
+    resources :appointments
+  end
+  get '/not_eligible', to: 'appointment_attempts#not_eligible'
```
benlovell: This should be moved into the `:appointment_attempts` resources declaration. Actually, it's worth checking but I think if you have a template with that name, you don't even need this routing declaration as they get rendered implicitly.

app/views/appointment_attempts/new.html.erb
AndrewVos: Yep! I normally do this but forgot

benlovell:
Just a reminder further to what I mentioned in slack - I think _ineligible_ is nicer than _not eligible_ in code and copy.

AndrewVos:
@benbarnett I renamed that link to `Book appointment` from `Create appointments`. We're using `Book appointment` in other places now too.

AndrewVos:
@benbarnett @benlovell Addressed all your comments.

PR #90, opened 2016-10-14
Don't configure gem 'biz' in an initializer

AndrewVos:
https://github.com/zendesk/biz/issues/18

This can be tested by going to `/appointments/new`.


Guntrisoft:
👍

app/models/business_days.rb
```diff
@@ -1,4 +1,17 @@
 class BusinessDays
+  # This can't happen in an initalizer because:
```
benlovell: Typo for initializer.

PR #92, opened 2016-10-17
Fix biz crashing on POST appointment_attempts/:id/appointments

AndrewVos:


app/models/business_days.rb
```diff
@@ -1,22 +1,26 @@
...
   end
 
   def self.before_now(amount)
+    configure
     Biz.time(amount, :day).before(Time.zone.now)
   end
+
+  # This can't happen in an initializer because:
+  # https://github.com/zendesk/biz/issues/18
+  def self.configure
```
benlovell: Eww. I'd move this to Puma's `on_worker_boot` hook instead.

PR #95, opened 2016-10-18
Guider can view their own calendar

benlovell:
This shows appointments scoped to the current user.

This ends at the edit appointment screen which remains to be implemented.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ ruby IO.read('.ruby-version').strip
 source 'https://rails-assets.org' do
   gem 'rails-assets-listjs'
   gem 'rails-assets-zloirock--core-js'
+  gem 'rails-assets-fullcalendar'
+  gem 'rails-assets-fullcalendar-scheduler'
+  gem 'rails-assets-qTip2'
 end
 
 source 'https://rubygems.org' do
@@ -16,7 +19,6 @@ source 'https://rubygems.org' do
   gem 'sprockets-es6'
   gem 'gds-sso'
   gem 'plek'
-  gem 'fullcalendar-rails'
   gem 'momentjs-rails'
   gem 'bootstrap-daterangepicker-rails'
   gem 'active_link_to'
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -13,6 +13,8 @@
 //= require turbolinks
 //= require moment
 //= require fullcalendar
+//= require fullcalendar-scheduler
+//= require qTip2
 //= require listjs
 //= require daterangepicker
 //= require core.js/core.js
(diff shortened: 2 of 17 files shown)


spec/features/guider_views_appointments_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    # this would appear 'today'
+    @appointment = create(:appointment, guider: current_user)
+    # this would appear 'tomorrow'
+    @tomorrow = create(:appointment, guider: current_user, start_at: BusinessDays.from_now(3))
+    # this won't appear for the current user
+    create(:appointment)
+  end
+
+  def when_they_view_their_calendar
+    @page = Pages::Calendar.new.tap(&:load)
```
benbarnett: image

app/assets/javascripts/calendars/guider-appointments.es6
```diff
@@ -0,0 +1,34 @@
+/* global Calendar, moment */
+{
+  'use strict';
+
+  class GuiderAppointmentsCalendar extends Calendar {
+    constructor(el, config = {}) {
+      const calendarConfig = $.extend(true, {
+        defaultDate: moment(el.data('default-date')),
```
Guntrisoft: Do we need to do this, as you have similar in base class `Calendar` ?

Can this line go into `Calendar`, so that defaultDate is always an instance of `moment` ?

Gemfile
```diff
@@ -3,6 +3,9 @@ ruby IO.read('.ruby-version').strip
 source 'https://rails-assets.org' do
   gem 'rails-assets-listjs'
   gem 'rails-assets-zloirock--core-js'
+  gem 'rails-assets-fullcalendar'
+  gem 'rails-assets-fullcalendar-scheduler'
+  gem 'rails-assets-qTip2'
```
Guntrisoft: Probably not an issue, but qTip isn't used here at all, but will be used once we get the company calendar in there....leave it up to you whether we leave it in or not.

app/assets/javascripts/application.js
```diff
@@ -13,13 +13,16 @@
 //= require turbolinks
 //= require moment
 //= require fullcalendar
+//= require fullcalendar-scheduler
+//= require qTip2
```
Guntrisoft: see above.

Gemfile
benlovell: It's the next thing I'm doing so may as well just leave it there.

AndrewVos:
This is a good pull request.

app/controllers/appointments_controller.rb
```diff
@@ -33,6 +44,13 @@ def show
 
   private
 
+  def date_range_params
+    starts = params[:start].to_date.beginning_of_day
+    ends   = params[:end].to_date.beginning_of_day
```
benlovell: It won't be missing unless the person hacks at the URL incorrectly. Also, that error doesn't get displayed in staging / production environments.

PR #96, opened 2016-10-18
Resource managers can view holidays

AndrewVos:
- Add a link to /holidays in the nav
- Appointments don't get assigned to guiders on a holiday
- Don't show bookable slots if there's a holiday
- Seed bank holidays with ActiveJob

### DIFF
--- a/app/assets/javascripts/calendars/holidays.es6
+++ b/app/assets/javascripts/calendars/holidays.es6
@@ -10,6 +10,7 @@
         columnFormat: 'ddd D/M',
         slotDuration: '00:30:00',
         eventBorderColor: '#000',
+        events: '/holidays',
         header: {
           'right': 'agendaWeek agendaThreeDay agendaDay today prev,next'
         },
@@ -22,16 +23,6 @@
       }, config);
 
       super(el, calendarConfig);
-
-      this.init();
-    }
-
-    handleEventData(data) {
-      this.$el.fullCalendar('addEventSource', data);
-    }
-
-    init() {
-      $.getJSON('/holidays', this.handleEventData.bind(this));
     }
   }
 
--- a/app/assets/javascripts/calendars/holidays.es6
+++ b/app/assets/javascripts/calendars/holidays.es6
@@ -1,0 +1,40 @@
+/* global Calendar */
+{
+  'use strict';
+
+  class HolidaysCalendar extends Calendar {
+    constructor(el, config = {}) {
+      const calendarConfig = $.extend(true, {
+        defaultView: 'agendaWeek',
+        weekends: false,
+        columnFormat: 'ddd D/M',
+        slotDuration: '00:30:00',
+        eventBorderColor: '#000',
+        header: {
+          'right': 'agendaWeek agendaThreeDay agendaDay today prev,next'
+        },
+        views: {
+          agendaThreeDay: {
+            type: 'agenda',
+            duration: { days: 3 }
+          }
+        }
+      }, config);
+
+      super(el, calendarConfig);
+
+      this.init();
+    }
+
+    handleEventData(data) {
+      this.$el.fullCalendar('addEventSource', data);
+    }
+
+    init() {
+      $.getJSON('/holidays', this.handleEventData.bind(this));
+    }
+  }
+
+  window.PWTAP = window.PWTAP || {};
+  window.PWTAP.HolidaysCalendar = HolidaysCalendar;
+}
--- a/app/controllers/holidays_controller.rb
+++ b/app/controllers/holidays_controller.rb
@@ -1,0 +1,12 @@
+class HolidaysController < ApplicationController
+  before_action :authorise_for_managing_resources!
+
+  def index
+  end
+
+  private
+
+  def authorise_for_managing_resources!
+    authorise_user!(User::RESOURCE_MANAGER_PERMISSION)
+  end
+end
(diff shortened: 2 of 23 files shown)


app/assets/javascripts/calendars/holidays.es6
```diff
@@ -0,0 +1,40 @@
...
+
+      this.init();
+    }
+
+    handleEventData(data) {
+      this.$el.fullCalendar('addEventSource', data);
+    }
+
+    init() {
+      $.getJSON('/holidays', this.handleEventData.bind(this));
```
Guntrisoft: So you should be able to use the `event` attribute of the full calendar config as a string to the 'API' end point where it can grab events

Guntrisoft:
Looks good to me, apart from the getting of the events in the JS

Branch needs a rebase too

benlovell:
Is this still WIP or can I review?

spec/models/appointment_spec.rb
```diff
@@ -105,6 +105,39 @@
           expect { subject.assign_to_guider }.to_not change { subject.guider }
         end
       end
+
+      context 'and the guider already has a holiday at that time' do
```
benbarnett: Do we want it like this? To me, it's fine for a guider to have multiple 'non working events' assigned to them. Don't think there's any harm in this really.

spec/models/appointment_spec.rb
```diff
@@ -105,6 +105,39 @@
...
+      context 'and there is a bank holiday at that time' do
+        before do
+          create(
+            :holiday,
+            start_at: appointment_start_time.beginning_of_day,
+            end_at: appointment_end_time.end_of_day
+          )
+        end
+
+        it 'does not assign to a guider' do
```
benbarnett: Same here re multiple non working events.

If a guider has 2 weeks off and there's a bank holiday in the middle of those 2 weeks, then it's fine to have both added?

spec/models/appointment_spec.rb
AndrewVos: This is about assigning Appointments to guiders, not Holidays :)

benbarnett: Oh yes, whoops.

app/models/holiday.rb
```diff
@@ -0,0 +1,14 @@
+class Holiday < ApplicationRecord
+  belongs_to :user, required: false
+
+  def self.merged_for_calendar_view
+    select(<<-SQL
+      DISTINCT ON(holidays.start_at, holidays.end_at, holidays.title)
+        holidays.title, holidays.start_at, holidays.end_at, holidays.title || COALESCE(' - ' || string_agg(users.name, ', '), '') AS title
+    SQL
```
benlovell: This indentation is a little off here.

app/models/holiday.rb
```diff
@@ -0,0 +1,14 @@
...
+          )
+      .joins('LEFT JOIN users ON users.id = holidays.user_id')
```
benlovell: I think `includes(:users)` would do the same here, plus the users would then be eager-loaded too as a bonus.

app/serializers/holiday_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class HolidaySerializer < ActiveModel::Serializer
+  attribute :title
+  attribute :start do
```
benlovell: You can use the `key` option for aliases instead and avoid yielding a block here and for `end` too.

config/routes.rb
```diff
@@ -17,6 +17,7 @@
   resources :customers
   resource :calendar, only: :show
   resources :appointments, only: %i(index show edit)
+  resources :holidays
```
benlovell: Can we start explicitly whitelisting the restful actions we support? It helps to understand what things we are supporting and where. Also cutting down the noise in `rails routes` is super helpful.

db/migrate/20161017130509_create_holidays.rb
```diff
@@ -0,0 +1,11 @@
+class CreateHolidays < ActiveRecord::Migration[5.0]
+  def change
+    create_table :holidays do |t|
+      t.integer  :user_id
```
benlovell: This needs an index.

app/controllers/holidays_controller.rb
```diff
@@ -0,0 +1,18 @@
+class HolidaysController < ApplicationController
+  before_action :authorise_for_managing_resources!
+
+  def index
+    respond_to do |format|
+      format.html
+      format.json do
+        render json: Holiday.merged_for_calendar_view
```
benlovell: This won't support the `start` and `end` dates provided by the calendar when the bindings (day, week, month, forward and backward) change, right?

AndrewVos: Just spoke about this. Answer is no, because we render all the events in one go.

AndrewVos:
@benlovell added tests https://github.com/guidance-guarantee-programme/telephone_appointment_planner/pull/96/commits/00886f875d5a9355a5743a5ac9a47a1338a2fc43

app/models/holiday.rb
AndrewVos: We don't really use users here though, the join is just so we can get at the user name.

app/models/holiday.rb
AndrewVos: I'm not actually sure how to get his right. Have a look at my screen when you get a moment

AndrewVos:
@benlovell all done here

PR #101, opened 2016-10-19
[do-not-merge] Fast bookable slots

AndrewVos:


AndrewVos:
Closing because we are now deleting and recreating all bookable slots, which is a bit faster.

PR #102, opened 2016-10-20
Improving file structure of scss files and adding scss-lint

Guntrisoft:
- moved each distinct 'component' into it's own scss file
- fixed linting issues
- added scss-lint to rake


Gemfile
```diff
@@ -45,6 +45,7 @@ source 'https://rubygems.org' do
     gem 'phantomjs-binaries'
     gem 'database_cleaner'
     gem 'chronic'
+    gem 'scss-lint', '~> 0.30'
```
Guntrisoft: stuck with the same version of the linter as pension_guidance, as I was having some issues running the rake task with the latest version of the linter installed

app/assets/stylesheets/application.scss
```diff
@@ -1,101 +1,9 @@
...
-
-.form-dob__input {
-  width: 64px;
-}
-
-.form-dob__input--year {
-  width: 90px;
-}
+@import "colours";
+@import "govuk_admin_template";
```
benbarnett: Shall we move these into `lib/` and `vendor/`?

Guntrisoft: Done 👍

PR #104, opened 2016-10-20
Make date of birth readonly after eligiblity check

benbarnett:
Before this, it was possible to provide an eligible date of birth
on the first page of the 'Book Appointment' journey, which is the
point at which the eligibility checks take place. However you
could then edit the date of birth to an ineligible age on the
second screen – resulting in an ineligible appointment being
successfully booked.

An unlikely scenario to occur, but this just prevents that from
happening.


Guntrisoft:
👍

AndrewVos:
This is good.

PR #105, opened 2016-10-20
Allow guiders to edit their own appointments

benlovell:
The guider can modify the personal details stored against appointments
assigned to them.

I extracted the common fields from new / edit into a partial to be
reused across both actions.

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -1,6 +1,6 @@
 class AppointmentsController < ApplicationController
-  before_action :authorise_for_agents!, except: %i(index edit)
-  before_action :authorise_for_guiders!, only: %i(index edit)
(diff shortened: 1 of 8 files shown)


AndrewVos:
How do they get to this page?

benlovell:
@AndrewVos from clicking an appointment on their calendar. There is a spec for that side of the behaviour over here: https://github.com/guidance-guarantee-programme/telephone_appointment_planner/blob/5ac65f702b1d6ccf15f4c81e13c64ae127a4339c/spec/features/guider_views_appointments_spec.rb#L13

Guntrisoft:
Looks good from my point of view 💯

AndrewVos:
This is the best pull request I have ever seen in my life. Please merge it.

PR #106, opened 2016-10-20
Use GDS-SSO supplied permissions check

benlovell:
GDS-SSO provides a predicate for us to check for the presence of
permissions.


Guntrisoft:
:+1:

PR #112, opened 2016-10-21
Refactor controller authentication to use blocks where possible

AndrewVos:
Less code == better code.


benlovell:
Rebase mate.

benlovell:
Definitely a fan of this style where possible. 💯

PR #115, opened 2016-10-21
Header corrections.

Guntrisoft:


benlovell:
We seem to have a mix of title / non title cased headers. Can we settle on one style?

PR #116, opened 2016-10-21
Making all page headers consistent

Guntrisoft:


benlovell:
Can we make the breadcrumbs follow suit?

benlovell:
Also, you missed https://github.com/guidance-guarantee-programme/telephone_appointment_planner/blob/2848cf085f02e5b1c344a91338c8374802cad6ed/app/views/schedules/edit.html.erb#L9

PR #119, opened 2016-10-21
Adding a status to appointments

Guntrisoft:
- only available on edit appointment screen
- defaults to 'pending' on creation of an appointment


benlovell:
Nice.

PR #123, opened 2016-10-22
Removes the time dependency from this spec

benlovell:
Many have tried, and failed.

cc @AndrewVos @Guntrisoft :poop:


benlovell:
That weekend self-merge.

PR #124, opened 2016-10-22
Use server clock to set current date

benlovell:
This allows us to time travel successfully in our feature specs and ends
a whole class of spec bugs caused by @AndrewVos' incompetence.

### DIFF
--- a/app/views/holidays/index.html.erb
+++ b/app/views/holidays/index.html.erb
@@ -7,4 +7,4 @@
   <h1>Holidays</h1>
 </div>
 
-<div data-module="HolidaysCalendar"></div>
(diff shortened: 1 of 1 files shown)


Guntrisoft:
😆👍 💯

AndrewVos:
Looks good, aside from obvious factual inaccuracies in the description (I am in fact a genius).

Might be better to just inject some javascript mocking the date when using `travel_to` though. This feels like a test concern bleeding into code.

benlovell:
@AndrewVos Maybe. But I can see the argument for both sides. Setting the default time / date using data attributes is certainly more straight-forward and it feels a bit scuzzy to be using mocks (javascript or otherwise) in feature specs.

AndrewVos:
Hm, I agree.

PR #125, opened 2016-10-24
Remove six week restriction on schedules

AndrewVos:
- Schedules have no six week restrictions anymore
- Destroy all bookable slots when regenerating
- Bookable slots are still generated for six weeks

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -6,7 +6,6 @@ class Schedule < ApplicationRecord
 
   validates :start_at, presence: true
   validates :start_at, uniqueness: { scope: :user_id }
-  validate :start_at_must_be_more_than_six_weeks_in_the_future, unless: :first_schedule?
 
   def self.active(day)
     order(:start_at)
@@ -28,7 +27,8 @@ class Schedule < ApplicationRecord
   end
 
   def modifiable?
-    start_at > 6.weeks.from_now
+    return true unless end_at
+    end_at > Time.zone.now.end_of_day
   end
 
   private
@@ -37,11 +37,5 @@ class Schedule < ApplicationRecord
     user
       .schedules
       .none?(&:persisted?)
-  end
-
-  def start_at_must_be_more_than_six_weeks_in_the_future
-    maximum_age = 6.weeks.from_now.beginning_of_day
-    older_than_six_weeks = start_at && start_at >= maximum_age
-    errors.add(:start_at, 'must be more than six weeks from now') unless older_than_six_weeks
   end
 end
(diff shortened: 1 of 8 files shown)


app/models/schedule.rb
```diff
@@ -27,21 +26,11 @@ def self.with_end_at
     select('*, LEAD(start_at) OVER (ORDER BY start_at) AS end_at')
   end
 
-  def modifiable?
```
benbarnett: Wondering if this should be just `end date is not in the past`?

app/models/schedule.rb
```diff
@@ -28,7 +27,7 @@ def self.with_end_at
   end
 
   def modifiable?
-    start_at > 6.weeks.from_now
+    (end_at.nil? || end_at > Time.zone.now.end_of_day)
```
benlovell: Maybe clearer with the `nil` check as a guard i.e.

```ruby
return true unless end_at

end_at > Time.zone.now.end_of_day
```

spec/factories/schedules.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :schedule do
     user { build(:user) }
-    start_at 6.weeks.from_now
+    start_at Time.zone.now.beginning_of_day
```
benlovell: This will be the same from the moment the factory is loaded. Is that the expected behaviour?

spec/features/resource_manager_manages_schedules_spec.rb
```diff
@@ -179,22 +170,22 @@ def and_the_guider_has_those_time_slots_available
...
   def and_the_schedule_is_deleted
     expect(@guider.reload.schedules).to be_empty
   end
 
-  def then_they_cant_delete_the_schedule
+  def then_they_cant_edit_the_schedule
     @page = Pages::EditUser.new
     @page.load(id: @guider.id)
-    expect(@page.delete).to be_disabled
+    expect(@page.schedules.first).to_not have_edit
```
benlovell: This is probably not the behaviour you want. I think you mean: `.to have_no_edit`. They are quite different semantically.

app/models/schedule.rb
AndrewVos: done

spec/factories/schedules.rb
AndrewVos: Fixed

spec/features/resource_manager_manages_schedules_spec.rb
AndrewVos: done

PR #126, opened 2016-10-24
Redirect home based on role

Guntrisoft:
- agents -> appointment_attempts/new
- guiders -> calendar
- resource managers -> guiders list (this should eventually be
  the resource calendar when it's been built)

### DIFF
--- a/spec/features/custom_role_redirects_spec.rb
+++ b/spec/features/custom_role_redirects_spec.rb
@@ -1,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.feature 'Custom Role Redirects' do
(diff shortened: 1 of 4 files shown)


spec/features/custom_role_redirects_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.feature 'Custom Role Redirects' do
+  context 'Resource Manager' do
```
benlovell: Can we collapse all of these contexts into the scenario title instead here?

Guntrisoft: Done!

PR #128, opened 2016-10-24
Improving UI for adding a new holiday

Guntrisoft:
- Moved new button to top left
- added breadcrumb to new holiday page
- fixed JS to make only X delete a holiday event

### DIFF
--- a/app/assets/javascripts/calendars/holidays.es6
+++ b/app/assets/javascripts/calendars/holidays.es6
@@ -11,23 +11,15 @@
         eventBorderColor: '#000',
         events: el.data('holidays-path'),
         header: {
-          center: 'title',
-          right: 'createHoliday month,agendaWeek,agendaDay today prev,next'
-        },
-        customButtons: {
-          createHoliday: {
-            text: 'Create holiday',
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/calendars/holidays.es6
```diff
@@ -11,23 +11,15 @@
...
-        eventRender: (event, element) => {
-          var $button = element.append(
-            '<button class="close t-delete-holiday">' +
-            '<span aria-hidden="true">X</span><span class="sr-only">Remove slot</span>' +
-            '</button>'
-          );
+        eventRender: (event, element, view) => {
+          const $button = $(`
+            <button class="close t-delete-holiday">
+              <span aria-hidden="true">X</span><span class="sr-only">Remove slot</span>
```
benbarnett: Shall we use glyphicon-remove here?

Guntrisoft: I'm thinking to leave this for now, as we are about to remove this way to delete things anyway - something to apply to the slot picker calendar though.

PR #131, opened 2016-10-24
Remove delete holiday spec as it's failing intermittently

Guntrisoft:
We will be altering the way you can delete an appointment soon anyway, so a completely new spec will replace this.


AndrewVos:
lgtm

PR #137, opened 2016-10-25
Fix orphaned bookable slots

AndrewVos:
There was a really irritating issue where sometimes we just couldn't
book appointments, randomly. This was because the bookable slot that
matched didn't have an existing user anymore, because we deleted that
user in db:seed.

The fix is to ensure we delete any child bookable slots when we delete
users.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -8,7 +8,7 @@ class User < ApplicationRecord
   default_scope { order(:name) }
 
   has_many :schedules, dependent: :destroy
-  has_many :bookable_slots
+  has_many :bookable_slots, dependent: :destroy, foreign_key: :guider_id
   has_many :appointments, foreign_key: :guider_id
   has_many :holidays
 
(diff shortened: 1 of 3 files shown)


Guntrisoft:
👍

benlovell:
Maybe this warrants a small spec, just to make sure this doesn't regress seeing it's pretty important behaviour?

AndrewVos:
Ok done dad @benlovell

app/models/user.rb
```diff
@@ -8,7 +8,7 @@ class User < ApplicationRecord
   default_scope { order(:name) }
 
   has_many :schedules, dependent: :destroy
-  has_many :bookable_slots
+  has_many :bookable_slots, dependent: :destroy, foreign_key: 'guider_id'
```
benlovell: Sorry to be a super-pest but can you provide the foreign key value as a symbol per the line below? 🕵️‍♀️

PR #141, opened 2016-10-25
Fix BookableSlots#with_holidays negating valid slots

AndrewVos:
Because of operator precedence we were dropping bookable slots for
users when the slots weren't actually covered by the holiday.


Guntrisoft:
👍

PR #143, opened 2016-10-26
Cancelled appointments are not considered for blocking bookable slots

Guntrisoft:
- if an appointment is cancelled then the bookable slot should be
  bookable

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -15,6 +15,7 @@ class BookableSlot < ApplicationRecord
     BusinessDays.from_now(3)
   end
 
+  # rubocop:disable Metrics/MethodLength
   def self.without_appointments
     joins(<<-SQL
             LEFT JOIN appointments ON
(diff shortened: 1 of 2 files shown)


app/models/bookable_slot.rb
```diff
@@ -23,7 +24,14 @@ def self.without_appointments
               appointments.end_at = #{quoted_table_name}.end_at
             SQL
          )
-      .where('appointments.start_at IS NULL')
+      .where(<<-SQL
```
benlovell:
```ruby
where(start_at: nil)
  .or(where(status: %i(cancelled_by_customer cancelled_by_pension_wise)))
```
Maybe that's a little clearer, although not sure it would work.

spec/models/bookable_slot_spec.rb
```diff
@@ -79,13 +79,14 @@ def result
...
-      context 'one guider has a bookable slot obscured by an appointment' do
+      context 'one guider has a bookable slot obscured by a non cancelled appointment' do
         it 'excludes the slot' do
           create(
             :appointment,
             guider: User.guiders.first,
             start_at: make_time(10, 30),
-            end_at: make_time(11, 30)
+            end_at: make_time(11, 30),
+            status: Appointment.statuses['pending']
```
benlovell: Pretty sure you can just pass the statuses as symbols and they get coerced correctly.

app/models/bookable_slot.rb
Guntrisoft: didn't work, and was struggling to get it to play ball with the or method (and having a join)

spec/models/bookable_slot_spec.rb
Guntrisoft: done!

PR #144, opened 2016-10-26
Show field update and message activities on appointments

AndrewVos:
It is important to note that ANY signed on user can edit an appointment now.

This is because we know who all these users are, and because the code would effectively allow any user role otherwise.

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -1,0 +1,40 @@
+class ActivitiesController < ApplicationController
+  def index
+    @activities = appointment.activities.since(since)
+
+    if @activities.present?
+      render partial: @activities
+    else
+      head :not_modified
+    end
+  end
+
+  def create
+    @activity = MessageActivity.create(message_params)
+
+    respond_to do |format|
+      format.js { render @activity }
+    end
+  end
+
+  private
+
+  def message_params
+    params
+      .require(:message_activity)
+      .permit(:message)
+      .merge(
+        user: current_user,
+        appointment: appointment
+      )
+  end
+
+  def since
+    timestamp = params[:timestamp].to_i / 1000
+    Time.zone.at(timestamp)
+  end
+
+  def appointment
+    Appointment.find(params[:appointment_id])
+  end
+end
--- a/app/views/activities/_activity_feed.html.erb
+++ b/app/views/activities/_activity_feed.html.erb
@@ -1,0 +1,37 @@
+<div class="row t-activity-feed">
+  <div class="col-md-12 border-bottom">
+    <h2>Recent activity</h2>
+
+    <div class="activity-feed">
+      <div class="well js-message-form">
+        <%= form_for MessageActivity.new(appointment: appointment), url: appointment_activities_path(appointment), remote: true do |f| %>
+          <div class="input-group">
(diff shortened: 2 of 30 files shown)


app/views/activities/_activity_feed.html.erb
```diff
@@ -0,0 +1,37 @@
+<div class="row t-activity-feed">
+  <div class="col-md-12 border-bottom">
+    <h2 class="h3">Recent activity</h2>
```
Guntrisoft: I think this can be a h2 size header on this page without looking odd

app/views/activities/_activity_feed.html.erb
```diff
@@ -0,0 +1,37 @@
...
+            <span class="input-group-btn">
+              <%= f.button class: 'btn btn-primary btn-block t-submit-message', data: { disable_with: 'Adding message...' } do %>
+                <span class="glyphicon glyphicon-plus" aria-hidden="true"></span>
+                Add message
+              <% end %>
+            </span>
+          </div>
+        <% end %>
+      </div>
+      <ul class="activity-feed__list"
```
Guntrisoft: At the moment this ends up rendering a line when there are no messages in the activity feed.

Would be great if we could have a `hide` class on the UL and then remove it when there are messages to show.

app/views/activities/_activity_feed.html.erb
```diff
@@ -0,0 +1,37 @@
...
+          <%= render(appointment.activities - Array(appointment.activities.first)) %>
+        </ul>
+        <button class="activity-feed__btn btn btn-default t-further-activities" data-target="#collapse" data-toggle="collapse" aria-expanded="false" aria-controls="collapse">
+          <span class="glyphicon glyphicon-th-list" aria-hidden="true"></span>
+          All activity
+        </button>
+      <% end %>
+    </div>
+  </div>
+</div>
```
Guntrisoft: can we put the `Fields marked with *Required are required.` underneath the `Personal Details` header

app/controllers/activities_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      )
+  end
+
+  def since
+    timestamp = params[:timestamp].to_i / 1000
+    Time.zone.at(timestamp)
+  end
+
+  def appointment
+    current_user.appointments.find(params[:appointment_id])
```
benlovell: The appointments will be scoped to `current_user` so this will break things in multiple ways here. I don't think we need to limit who can comment on what outside of any other enforced permissions so I reckon we could just drop the scoping here.

app/views/activities/_activity_feed.html.erb
AndrewVos: Done

app/controllers/activities_controller.rb
AndrewVos: Totally missed that. Thanks, done.

app/views/activities/_activity_feed.html.erb
AndrewVos: How should I do this do you think? I tried `:empty` but then remembered that the element has to be proper empty :(

Guntrisoft: I was thinking that the UL could have a hide class on initial render and then at any point in the code where an activity feed item gets added then remove the hide class on the UL

PR #146, opened 2016-10-26
Ensure that calendar stays on current view/date across page impressions

Guntrisoft:
- Storing a cookie of the current view and date of each calendar
- Config for fullcalendar on initialisation gets overriden with values from cookie
- Remembers current defaultView/defaultDate settings for a year (across browser sessions)

### DIFF
--- a/app/assets/javascripts/calendar.es6
+++ b/app/assets/javascripts/calendar.es6
@@ -1,27 +1,58 @@
-/* global moment */
+/* global moment, cookie */
 
 'use strict';
 
+window.PWTAP = window.PWTAP || {};
+
 class Calendar {
   constructor(el, config = {}) {
+    this.$el = el;
+
     const defaultConfig = {
-      schedulerLicenseKey: 'GPL-My-Project-Is-Open-Source',
-      weekends: false,
-      timeFormat: 'H:mm',
-      slotLabelFormat: 'H:mm',
+      allDaySlot: false,
+      cookieName: this.$el.attr('id') || 'calendar',
+      defaultDate: moment(this.$el.data('default-date')),
+      firstDay: 1,
+      height: 'auto',
       maxTime: '19:00:00',
       minTime: '08:30:00',
-      height: 'auto',
-      allDaySlot: false,
-      firstDay: 1,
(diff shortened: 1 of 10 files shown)


Guntrisoft:
@benbarnett

app/assets/javascripts/calendar.es6
```diff
@@ -14,14 +17,44 @@ class Calendar {
       height: 'auto',
       allDaySlot: false,
       firstDay: 1,
-      defaultDate: moment(el.data('default-date'))
+      defaultDate: moment(el.data('default-date')),
+      viewRender: (view) => {
```
benbarnett: The issue with having the full method in the default config is that this'll get blown away if a `viewRender` is passed in with the config.

Might be worth splitting this out, calling the 'default' one, and then running the passed in `config.viewRender` after the fact (if defined).

app/assets/javascripts/calendar.es6
```diff
@@ -1,9 +1,12 @@
-/* global moment */
+/* global moment, cookie */
 
 'use strict';
 
 class Calendar {
   constructor(el, config = {}) {
+    this.cookieName = this.constructor.name;
```
benbarnett: Worth having this as part of the config? In the case of multiple calendars on a page etc.

AndrewVos:
This lgtm, though I wonder if `defaultDate` should be stored. What if the user is a few months ahead? It might be weird coming back to the calendar and being in the wrong month?

Guntrisoft:
@AndrewVos I can see some negatives and positives to storing it!

If they are editing an appointment on a particular date in the future and come back to it, then losing where they were date wise might be annoying.

Maybe see how they go with it, and whether TPAS/TP note any issues?

Guntrisoft:
@benbarnett this has evolved a bit since the last time you reviewed it, based on the feedback you gave.

benlovell:
@Guntrisoft @AndrewVos We definitely need some means of directing them back to the right day / view of the calendar after batch modifying a bunch of appointments. Whether it's via this approach or not I don't know...

AndrewVos:
Agree with that @Guntrisoft.

LGTM

PR #148, opened 2016-10-27
Agents can search for appointments

AndrewVos:
- Search on #first_name, #last_name, and #id.
- Search on a date range
- Adds pg_search

NOTE: There are no `tsvector` indexes yet!

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -54,14 +54,18 @@ class AppointmentsController < ApplicationController
   end
 
   def search
-    search_params = params[:search] || {}
     @search = Search.new(
       search_params[:q],
       search_params[:date_range]
     )
+    @results = @search.results.page(params[:page])
   end
 
   private
+
+  def search_params
+    params.fetch(:search, {}).permit(:q, :date_range)
+  end
 
   def date_range_params
     starts = params[:start].to_date.beginning_of_day
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -53,6 +53,14 @@ class AppointmentsController < ApplicationController
     @appointment = Appointment.find(params[:id])
   end
 
+  def search
+    search_params = params[:search] || {}
+    @search = Search.new(
+      search_params[:q],
+      search_params[:date_range]
+    )
+  end
+
   private
 
   def date_range_params
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -1,4 +1,8 @@
 class Appointment < ApplicationRecord
+  include PgSearch
+
+  pg_search_scope :search, against: [:id, :first_name, :last_name]
+
   enum status: %i(
     pending
     completed
@@ -26,6 +30,17 @@ class Appointment < ApplicationRecord
   has_many :activities, -> { order('created_at DESC') }
 
   audited on: :update
+
+  def self.full_search(query, start_at, end_at)
+    results = all
+    results = results.search(query) if query.present?
+    if start_at && end_at
+      results = results
+                .where('start_at > ?', start_at)
+                .where('end_at < ?', end_at)
+    end
+    results
+  end
 
   def assign_to_guider
     slot = BookableSlot
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -1,7 +1,7 @@
 class Appointment < ApplicationRecord
   include PgSearch
 
-  pg_search_scope :search, against: [:id, :first_name, :last_name]
+  pg_search_scope :search, against: %i(id first_name last_name)
 
   enum status: %i(
     pending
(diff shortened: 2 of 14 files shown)


app/controllers/appointments_controller.rb
```diff
@@ -53,6 +53,15 @@ def show
     @appointment = Appointment.find(params[:id])
   end
 
+  def search
+    search_params = params[:search] || {}
```
benlovell: Can you move this to something like this?

```ruby
def search_params
  params.fetch(:search, {}).permit(:query, :date_range)
end
```

app/models/appointment.rb
```diff
@@ -1,4 +1,8 @@
 class Appointment < ApplicationRecord
+  include PgSearch
+
+  pg_search_scope :search, against: [:id, :first_name, :last_name]
```
benlovell: Can you use `%i(...)` instead of the symbol array? It's consistent elsewhere then.

app/views/appointments/search.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  <ol class="breadcrumb">
+    <li><a href="<%= root_path %>">Home</a></li>
+    <li class="active">Search</li>
+  </ol>
+  <h1>Search</h1>
+</div>
+
+<div class="row form-group">
+  <div class="col-md-12">
+    <%= form_for @search, layout: :basic, method: 'GET', url: {action: :search} do |f| %>
```
benlovell: Can you add some spaces in those hash braces?

app/views/appointments/search.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @results.each do |result| %>
+      <tr class='t-result'>
+        <td class='t-id'><%= result.id %></td>
+        <td><%= result.first_name %></td>
+        <td><%= result.last_name %></td>
+        <td><%= result.start_at %></td>
+        <td><%= result.status %></td>
```
benlovell: We should `titleize` this.

app/views/appointments/search.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <th>Status</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @results.each do |result| %>
+      <tr class='t-result'>
+        <td class='t-id'><%= result.id %></td>
+        <td><%= result.first_name %></td>
+        <td><%= result.last_name %></td>
+        <td><%= result.start_at %></td>
```
benlovell: This should to be formatted.

config/routes.rb
```diff
@@ -18,6 +18,7 @@
   resource :calendar, only: :show
   resources :appointments, only: %i(index show edit update) do
     resources :activities, only: %i(index create)
+    get '/search', on: :collection, action: :search
```
benlovell: The `action` is superfluous here I think. It would be inferred.

spec/features/agent_searches_for_appointments_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.feature 'Agent searches for appointments' do
+  before do
```
benlovell: This makes more sense when expressed in something like `def and_appointments_exist` that you refer to in our already given-when-then style specs.

spec/features/agent_searches_for_appointments_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  def then_they_can_see_only_that_result
+    expect(@page).to have_results(count: 1)
+    expect(@page.results.first.id.text.to_i).to eq @expected_appointment.id
+  end
+
+  def and_there_is_an_appointment_in_the_future
+    @expected_appointment = create(
+      :appointment,
+      start_at: 20.days.from_now,
+      end_at: 20.days.from_now + 1.hour
```
benlovell: This is inferred from `start_at` right? We can drop it.

config/routes.rb
AndrewVos: Can't remove that

spec/features/agent_searches_for_appointments_spec.rb
AndrewVos: Yup

app/models/appointment.rb
AndrewVos: Done

app/views/appointments/search.html.erb
AndrewVos: Done

app/views/appointments/search.html.erb
AndrewVos: Done

app/views/appointments/search.html.erb
AndrewVos: Done

PR #153, opened 2016-10-28
Refactor fullcalender appointment booker to use XHR

AndrewVos:


config/routes.rb
```diff
@@ -24,6 +24,9 @@
   resources :holidays, only: %i(index new create) do
     delete '/', on: :collection, action: :destroy
   end
+  resources :bookable_slots, only: [] do
```
benlovell: `get '/bookable_slots/available', to: 'bookable_slots#available'` makes more sense here I think.

app/views/appointments/new.html.erb
```diff
@@ -12,7 +12,13 @@
     <div class="col-md-12">
       <h2>Choose available slot <%= required_label %></h2>
       <p class="text-danger hide" id="slot-error">Please choose a slot</p>
-      <div class="appointment-availability-calendar" data-module="AppointmentAvailabilityCalendar" data-error-id="slot-error" data-events=".slots-data" data-default-date="<%= BookableSlot.next_valid_start_date %>"></div>
+      <div
+          class="appointment-availability-calendar"
+          data-module="AppointmentAvailabilityCalendar"
+          data-error-id="slot-error"
+          data-available-slots-path="<%= available_bookable_slots_path%>"
```
benlovell: Missing a space at the end.

app/controllers/bookable_slots_controller.rb
```diff
@@ -0,0 +1,7 @@
+class BookableSlotsController < ApplicationController
+  def available
```
benlovell: I wonder if this makes more sense as `index`?

PR #154, opened 2016-10-29
Appointment notifications

benlovell:
Sends customer / guider notifications by email and pusher respectively.

For now this is only triggered when the resource manager batch updates appointments. I'll sneak around separately and ensure notifications are triggered in the correct places.

Details are in the commits. I'm going to tidy up the date formatting separately so ignore that for the time being please.

### DIFF
--- a/app/views/calendars/show.html.erb
+++ b/app/views/calendars/show.html.erb
@@ -7,4 +7,38 @@
   <h1>My appointments</h1>
 </div>
 
-<div id="GuiderAppointmentsCalendar" data-module="GuiderAppointmentsCalendar" data-default-date="<%= Time.zone.now %>"></div>
+<div
+  data-module="GuiderAppointmentsCalendar"
+  data-default-date="<%= Time.zone.now %>"
+  data-guider-id="<%= current_user.id %>"
+></div>
+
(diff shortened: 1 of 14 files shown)


app/views/calendars/show.html.erb
```diff
@@ -7,4 +7,38 @@
...
+      </div>
+      <div class="modal-body"></div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+      </div>
+    </div>
+  </div>
+</div>
+
+<% content_for :body_end do %>
```
AndrewVos: Can we move this out of the view?

config/environments/test.rb
```diff
@@ -41,6 +41,9 @@
   # Precompile in travis to resolve poltergeist timeouts
   config.assets.compile = ENV['TRAVIS'] ? false : true
 
+  # work off jobs inline
+  config.active_job.queue_adapter = :inline
```
AndrewVos: You can probably delete those `perform_enqueued` called now?

AndrewVos:
This looks good.. Rebase and merge :)

config/environments/test.rb
benlovell: Scum bag.

PR #158, opened 2016-10-31
Adding method to return full name of appointment

Guntrisoft:
- output the full name in breadcrumb and title to help
  user know what they are edit/rescheduling

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -42,6 +42,10 @@ class Appointment < ApplicationRecord
     results
   end
 
+  def name
+    "#{first_name} #{last_name}"
+  end
+
   def assign_to_guider
     slot = BookableSlot
            .without_appointments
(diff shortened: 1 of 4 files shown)


spec/models/appointment_spec.rb
```diff
@@ -187,4 +187,14 @@ def results(q, start_at, end_at)
       expect(results).to eq [appointment]
     end
   end
+
+  describe '#name' do
+    let(:subject) do
```
benlovell: You can just use `subject { ... }` here instead.

app/models/appointment.rb
```diff
@@ -42,6 +42,10 @@ def self.full_search(query, start_at, end_at)
     results
   end
 
+  def name
+    [first_name, last_name].join(' ')
```
benlovell: Interpolating this would be more concise.

spec/models/appointment_spec.rb
Guntrisoft: done

app/models/appointment.rb
Guntrisoft: done

benlovell:
:100:

PR #159, opened 2016-10-31
Consolidate navigation into dropdown menus

benbarnett:
We were running out of room a little bit.

Also adds a link to the resource manager calendar if applicable. The labels could probably be improved but it looks less broken for now.

screen shot 2016-10-31 at 14 28 32


benlovell:
Nice 👍

PR #160, opened 2016-10-31
Resource managers see holidays as background events in resource calendar

AndrewVos:

### DIFF
--- a/app/controllers/holidays_controller.rb
+++ b/app/controllers/holidays_controller.rb
@@ -1,11 +1,17 @@
 class HolidaysController < ApplicationController
   before_action :authorise_for_resource_managers!
 
+  def merged
+    render json: Holiday.merged_for_calendar_view
+  end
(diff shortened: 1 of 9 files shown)


benlovell:
Can we get some coverage of this in a feature spec? Just to be sure that the calendar bits all hook up and don't regress.

app/controllers/holidays_controller.rb
```diff
@@ -1,11 +1,17 @@
 class HolidaysController < ApplicationController
   before_action :authorise_for_resource_managers!
 
+  def merged
```
benlovell: Am I missing something or is this not actually being used?

benlovell: Wait, ignore that.

AndrewVos: It's being used for `holidays#index`

PR #161, opened 2016-11-01
Display a calendar for guiders which shows all company appointments

Guntrisoft:
screen shot 2016-11-01 at 10 20 29

### DIFF
--- a/spec/features/guider_views_appointments_spec.rb
+++ b/spec/features/guider_views_appointments_spec.rb
@@ -15,6 +15,19 @@ RSpec.feature 'Guider views appointments' do
     end
   end
 
+  scenario 'Guider views all appointments', js: true do
+    given_the_user_is_a_guider do
+      and_there_are_appointments_for_multiple_guiders
+      travel_to @appointment.start_at do
+        when_they_view_the_company_calendar
+        then_they_see_the_appointments_for_today
+        when_they_advance_a_working_day
+        then_they_see_the_appointments_for_that_day
+        and_they_cannot_edit_the_appointment
+      end
+    end
+  end
+
   def and_there_are_appointments_for_multiple_guiders
     # this would appear 'today'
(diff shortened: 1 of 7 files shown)


spec/features/roles_spec.rb
```diff
@@ -155,4 +171,9 @@ def when_they_try_to_reschedule_an_appoinment
     @page = Pages::RescheduleAppointment.new
     @page.load(id: appointment.id)
   end
+
+  def when_they_try_to_view_the_company_calendar
+    @page = Pages::CompanyCalendar.new
```
AndrewVos: `@page = Pages::CompanyCalendar.new.tap&(:load)` in before ben

spec/features/guider_views_appointments_spec.rb
```diff
@@ -40,6 +57,11 @@ def then_they_see_the_appointment_for_today
     end
   end
 
+  def then_they_see_the_appointments_for_today
+    @page.wait_for_appointments
+    expect(@page).to have_appointments(count: 1)
```
AndrewVos: I would test that he appointment title is displayed, otherwise we're not sure the "next" button click did anything really.

spec/features/roles_spec.rb
benlovell: 😄

spec/features/guider_views_appointments_spec.rb
Guntrisoft: On this calendar there isn't any titles shown inside the event elements - only the qtips when you hover over - I'll try and come up something though.

benlovell: Maybe just tag it with the ID of the appointment it's bound to and assert on the presence of that?

Guntrisoft: Done !

PR #164, opened 2016-11-01
add in jump to date button to calendars where it is relevant

Guntrisoft:
screen shot 2016-11-01 at 15 46 59

### DIFF
--- a/app/assets/javascripts/calendar.es6
+++ b/app/assets/javascripts/calendar.es6
@@ -11,6 +11,12 @@ class Calendar {
     const defaultConfig = {
       allDaySlot: false,
       cookieName: this.$el.attr('id') || 'calendar',
+      customButtons: {
+        jumpToDate: {
+          text: 'Jump to date',
+          click: this.jumpToDateClick.bind(this)
+        }
+      },
       defaultDate: moment(this.$el.data('default-date')),
       firstDay: 1,
       height: 'auto',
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/calendar.es6
```diff
@@ -11,6 +11,21 @@ class Calendar {
     const defaultConfig = {
       allDaySlot: false,
       cookieName: this.$el.attr('id') || 'calendar',
+      customButtons: {
+        jumpToDate: {
+          text: 'Jump to date',
+          click: (event) => {
+            this.$jumpToDateEl.val(this.getCurrentDate());
+
+            const dateRangePicker = this.$jumpToDateEl.data('daterangepicker');
```
benbarnett: Can we try to keep the variable declarations at the top of each method?

app/assets/javascripts/calendar.es6
```diff
@@ -31,6 +46,34 @@ class Calendar {
     );
 
     this.$el.fullCalendar(this.config);
+
+    this.insertJumpToDate();
+  }
+
+  insertJumpToDate() {
+    const jumpToDateButton = $('.fc-jumpToDate-button');
```
benbarnett: Should we prefix with `$` as it's a jQuery object?

AndrewVos:
Looks good. Wondering if tests are worth the effort?

app/assets/javascripts/calendar.es6
```diff
@@ -11,6 +11,21 @@ class Calendar {
     const defaultConfig = {
       allDaySlot: false,
       cookieName: this.$el.attr('id') || 'calendar',
+      customButtons: {
+        jumpToDate: {
+          text: 'Jump to date',
+          click: (event) => {
```
benbarnett: I know we don't have tests yet but if this was moved into it's own named method it'll make it easier to test & debug when we do.

app/assets/javascripts/calendar.es6
```diff
@@ -11,6 +11,21 @@ class Calendar {
...
+            this.$jumpToDateEl.val(this.getCurrentDate());
+
+            const dateRangePicker = this.$jumpToDateEl.data('daterangepicker');
+
+            dateRangePicker.setStartDate(this.getCurrentDate());
+            dateRangePicker.setEndDate(this.getCurrentDate());
```
benbarnett: Worth declaring `this.getCurrentDate()` as a variable as we're referencing it 3 times?

app/assets/javascripts/calendar.es6
Guntrisoft: done

app/assets/javascripts/calendar.es6
Guntrisoft: done

PR #166, opened 2016-11-01
Display an error when failing to assign an appointment to a guider

AndrewVos:
~WIP until pete merges all his deletions~

Also rerenders `reschedule` page when a user did not choose a slot.


Guntrisoft:
Looks good from my point of view

PR #168, opened 2016-11-02
Rejigged how full calendar methods are defined

Guntrisoft:
- follows pattern already used with viewRender method
- instead of passing in methods as part of the defaultConfig
  this moves them to methods in the class itself
- base class proxies fullcalendar event calls


app/assets/javascripts/calendar.es6
```diff
@@ -25,8 +25,15 @@ class Calendar {
       schedulerLicenseKey: 'GPL-My-Project-Is-Open-Source',
       slotLabelFormat: 'H:mm',
       timeFormat: 'H:mm',
+      weekends: false,
+      select: (start) => this.select(start),
```
benbarnett: Does `() => this.select(arguments)` work?

PR #169, opened 2016-11-02
Deliver appointment confirmation email

AndrewVos:

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: 'Pension Wise Appointments <appointments@pensionwise.gov.uk>'
   layout 'mailer'
 end
--- a/spec/features/agent_manages_appointments_spec.rb
+++ b/spec/features/agent_manages_appointments_spec.rb
@@ -10,6 +10,16 @@ RSpec.feature 'Agent manages appointments' do
       when_they_want_to_book_an_appointment
       and_they_fill_in_their_appointment_details
       then_that_appointment_is_created
+      and_the_customer_gets_an_email_confirmation
+    end
+  end
+
+  scenario 'Agent creates appointment without an email' do
+    given_the_user_is_an_agent do
+      and_there_is_a_guider_with_available_slots
+      when_they_want_to_book_an_appointment
+      and_they_fill_in_their_appointment_details_without_an_email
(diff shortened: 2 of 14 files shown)


spec/mailers/appointment_mailer_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        expect(body).to include('12:00am, 23 October 2016')
+        expect(body).to include("reference number, ##{appointment.id}")
+      end
+
+      it 'includes the guider' do
+        expect(body).to include(appointment.guider.name)
+      end
+    end
+  end
+  # it 'renders the subject' do
```
Guntrisoft: ?

AndrewVos: hehe, blatantly copying and pasting from `planner`

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: 'appointments@pensionwise.gov.uk'
```
benlovell: Can we add the sender name here too? In the form `Pension Wise Appointments <appointments@pensionwise.gov.uk>`

spec/features/agent_manages_appointments_spec.rb
```diff
@@ -10,6 +10,18 @@
       when_they_want_to_book_an_appointment
       and_they_fill_in_their_appointment_details
       then_that_appointment_is_created
+      and_the_customer_gets_an_email_confirmation
+    end
+  end
+
+  context 'Customer does not give an email' do
```
benlovell: Can we flatten these contexts? When they only have a single context (and in a feature-style spec) there isn't any point separating them.

app/mailers/application_mailer.rb
AndrewVos: done

spec/features/agent_manages_appointments_spec.rb
AndrewVos: done

benlovell:
You'll also need to add: `config.action_mailer.default_url_options = { host: ENV['APPLICATION_HOST'] }` to `production.rb`

benlovell:
Also I forgot to mention previously but we have text templates in planner. We should over here too.

AndrewVos:
@benlovell done and done

benlovell:
Lovely 💯

PR #170, opened 2016-11-02
Ensure that list js ordering of appointments in search works

Guntrisoft:
- ordering the search page by default by created_at
- when there is a search it orders by relevance
- added created_at as a column for clarity


AndrewVos:
We should really start storing our date formats in a yaml or something at some point :cry:

AndrewVos:
LGTM

PR #174, opened 2016-11-02
Fix broken tests

AndrewVos:
Not sure how they ever really worked?


benlovell:
Can you just cherry-pick 60aa880 as that makes sense, the other commit is pure madness and I don't ever want to see it again. 😄

AndrewVos:
okie merge pls

PR #175, opened 2016-11-03
Sticky header for resource calendar

Guntrisoft:
- Makes the calendar as large as the viewport
- Ensures that names of guiders remain on the screen as you scroll down

### DIFF
--- a/app/assets/javascripts/calendars/guiders-appointments.es6
+++ b/app/assets/javascripts/calendars/guiders-appointments.es6
@@ -30,10 +30,38 @@
 
       super(el, calendarConfig);
 
+      this.alterHeight();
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/calendars/guiders-appointments.es6
```diff
@@ -30,12 +30,40 @@
...
+      this.alterHeight();
+      $(window).on('resize', this.debounce(this.alterHeight.bind(this), 20));
+
       this.eventChanges = [];
       this.actionPanel = $('[data-action-panel]');
 
       this.setupUndo();
     }
 
+    debounce(func, wait, immediate) {
```
benbarnett: Feels like this could be promoted up but I guess let's see how much it gets used

PR #177, opened 2016-11-03
Notify customer when Agent reschedules an appointment

AndrewVos:


app/mailers/appointment_mailer.rb
```diff
@@ -3,4 +3,9 @@ def confirmation(appointment)
     @appointment = appointment
     mail to: @appointment.email, subject: 'Your Pension Wise Appointment'
   end
+
+  def updated(appointment)
+    @appointment = appointment
+    mail to: @appointment.email, subject: 'Your Pension Wise Appointment'
```
benlovell: You can specify the `subject` option at the mailer level to remove some of this duplication.

PR #180, opened 2016-11-03
Ensure calendar doesn't break when guider for an event doesn't exist

Guntrisoft:
- in the odd situation where there is an event without a guider
  in the resource list this ensures that the calendar doesn't
  break when in timeline mode

### DIFF
--- a/app/assets/javascripts/calendars/guiders-appointments.es6
+++ b/app/assets/javascripts/calendars/guiders-appointments.es6
@@ -77,7 +77,7 @@
         return;
       }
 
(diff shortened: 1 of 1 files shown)


benbarnett:
Makes me slightly nervous this one! I guess this might happen if guiders have since left the organisation? Are we seeing this issue anywhere else?

Guntrisoft:
It's certainly a scenario that needs thinking through thoroughly - it came about because of a problem in factory created data where it allocates appointments to any type of user

PR #184, opened 2016-11-04
Rewrite JS modules to use GOVUK Admin template module loader

Guntrisoft:
- every module uses TapBase class which deals with setting up config
- all modules can now load config from a cookie
- figured out a way to trigger an event drop in fullcalendar
  without needing to access the instance of the module
  (which we no longer have access to now that we are using
  GOVUK Admin template module loader)

### DIFF
--- a/spec/support/pages/resource_calendar.rb
+++ b/spec/support/pages/resource_calendar.rb
@@ -44,7 +44,7 @@ module Pages
           #{yield}
 
           calendar.fullCalendar('updateEvent', event);
-          calendar.data('loaded-module').handleEventChange(event, function() {});
+          calendar.data('fullCalendar').view.triggerEventDrop(event, moment(), function() {}, $('.fc-event'));
         }();
       JS
     end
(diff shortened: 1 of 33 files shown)


spec/support/pages/resource_calendar.rb
```diff
@@ -15,14 +15,14 @@ def reschedule(appointment, hours:, minutes:)
...
-          var event = calendar.fullCalendar('clientEvents', #{id})[0];
+          var calendar = $('.js-calendar'),
+          event = calendar.fullCalendar('clientEvents', #{id})[0];
 
           event.start.hours(#{hours}).minutes(#{minutes});
           event.end.hours(#{hours + 1}).minutes(#{minutes});
 
           calendar.fullCalendar('updateEvent', event);
-          calendar.data('loaded-module').handleEventChange(event, function() {});
+          calendar.data('fullCalendar').view.triggerEventDrop(event, moment(), function() {}, $('.fc-event'));
```
benlovell: Nice.

benlovell:
Beat you to the merge / rebase. 😄

PR #189, opened 2016-11-06
Simplify appointment update

benlovell:
We didn't need to tap into the appointment, `.update` has the same
semantics.


Guntrisoft:
👍

PR #190, opened 2016-11-06
Return slots deterministically

benlovell:
This should resolve the intermittent test failures (the latest ones, at
least) in the CI environment. It seems the specs assert on the
particular ordering of `Schedule#slots` but the order was not
guaranteed.

I added a default scope as it's easiest and it seemed to make sense to
order slots in a schedule by their ordinal day of week.


benlovell:
This is getting the self-merge treatment so I can see if it passes in CI.

PR #193, opened 2016-11-07
Push notify for appointments on the same day

benlovell:
Limits the guider notifications to the same day otherwise the guider
may be overwhelmed by notifications and do a frighten.

### DIFF
--- a/app/jobs/pusher_notification_job.rb
+++ b/app/jobs/pusher_notification_job.rb
@@ -2,6 +2,8 @@ class PusherNotificationJob < ApplicationJob
   queue_as :default
 
   def perform(guider_id, appointment)
+    return unless current?(appointment)
+
     Pusher.trigger(
       'telephone_appointment_planner',
       guider_id.to_s,
@@ -9,5 +11,11 @@ class PusherNotificationJob < ApplicationJob
       start: appointment.start_at.strftime('%d %B %Y %H:%M'),
       guider_name: appointment.guider.name
     )
+  end
+
+  private
+
+  def current?(appointment)
+    appointment.start_at.to_date == Time.zone.today.to_date
   end
 end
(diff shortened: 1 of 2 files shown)


Guntrisoft:
👍

spec/jobs/pusher_notification_job_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe PusherNotificationJob, '#perform' do
+  subject { described_class.new.perform(0, appointment) }
```
AndrewVos: Can we use nil here instead? `0` seems like a magic number to me. Bit confusing

AndrewVos:
Wow, commented an hour late! Never mind that then.

spec/jobs/pusher_notification_job_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  context 'when the appointment is for today' do
+    let(:appointment) { build_stubbed(:appointment, start_at: Time.zone.today) }
```
dwhenry: `Time.zone.today` returns as date while start_at should be a  datetime. Can we use `Time.zone.now` instead

spec/jobs/pusher_notification_job_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    it 'sends the push notification' do
+      expect(Pusher).to receive(:trigger)
+
+      subject
+    end
+  end
+
+  context 'when the appointment is for another day' do
+    let(:appointment) { build_stubbed(:appointment, start_at: Time.zone.tomorrow) }
```
dwhenry: same here with `1.day.from_now` instead

app/jobs/pusher_notification_job.rb
```diff
@@ -10,4 +12,10 @@ def perform(guider_id, appointment)
       guider_name: appointment.guider.name
     )
   end
+
+  private
+
+  def current?(appointment)
+    appointment.start_at.to_date == Time.zone.today.to_date
```
dwhenry: as `Time.zone.today` is a date we should be able to drop the `.to_date`

PR #194, opened 2016-11-07
Add assigned and reassigned activities to appointments

AndrewVos:
I'm tempted to introduce an additional STI called `ReassignmentActivity` to simplify things.

Also, notice that `Activity` now has `owner` and `prior_owner`. They will be used to query activities when we have a per-user activity-feed.

### DIFF
--- a/db/migrate/20161107140328_add_prior_owner_id_to_activities.rb
+++ b/db/migrate/20161107140328_add_prior_owner_id_to_activities.rb
@@ -1,0 +1,5 @@
+class AddPriorOwnerIdToActivities < ActiveRecord::Migration[5.0]
+  def change
+    add_column :activities, :prior_owner_id, :integer, index: true
+  end
+end
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -1,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe Activity, type: :model do
+  describe 'validations' do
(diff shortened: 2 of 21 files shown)


db/migrate/20161107140328_add_prior_owner_id_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddPriorOwnerIdToActivities < ActiveRecord::Migration[5.0]
+  def change
+    add_column :activities, :prior_owner_id, :integer, index: true
```
benlovell: Any particular reason these migrations weren't combined? Not a big deal, just wondering?

AndrewVos: I kinda like using `:e db/migrate/*owner_id*<tab>` to browse to files in vim... That's all really.

spec/models/activity_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    it 'is valid with valid parameters' do
+      expect(subject).to be_valid
+    end
+
+    it 'validates presence of owner' do
+      subject.owner = nil
+      subject.validate
+      expect(subject.errors[:owner]).to_not be_empty
+    end
```
dwhenry: We have used the `shoulda-matchers` to do this sort of validation checking in other projects.

PR #202, opened 2016-11-08
Make default new schedule start date 'today'

benbarnett:
Since removing the 6 week window restrictions around creating
and modifying schedules, the previous default start date no
longer made sense.


Guntrisoft:
👍

PR #205, opened 2016-11-09
Alerts the resource manager that they are about to lose changes

Guntrisoft:
- if they move/resize an appointment then they are warned
  they are going to lose changes if they then try and navigate
  away
- enable links through to edit appointment page for events
  on resource manager calendar and company calendar for a guider


benlovell:
We'll need some coverage for the cancellation stuff too. Happy to do this if you want me to?

Guntrisoft:
@benbarnett split it into two

PR #208, opened 2016-11-09
Disable cookie config saving when using resource calendar

Guntrisoft:
- Store event changes with XHR
- Ensures that resource calendar always defaults to today
  when returning back to the calendar
- Ensures that the company calendar is unaffected by this
  change

### DIFF
--- a/spec/features/resource_manager_modifies_appointments_spec.rb
+++ b/spec/features/resource_manager_modifies_appointments_spec.rb
@@ -184,15 +184,6 @@ RSpec.feature 'Resource manager modifies appointments' do
     @other_appointment = create(:appointment, guider: @jan)
   end
 
-  def then_they_see_todays_date_shown
-    @page.wait_until_date_visible
-    expect(@page.date.text).to eq Time.zone.now.strftime('%B %e, %Y')
-  end
-
-  def when_they_click_the_next_day_button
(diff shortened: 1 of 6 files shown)


spec/features/resource_manager_modifies_appointments_spec.rb
```diff
@@ -138,7 +137,7 @@ def and_there_is_a_bookable_slot_for_a_guider
     )
   end
 
-  def then_they_can_see_the_bookable_slot
+  def then_they_can_see_the_bookable_slot # rubocop:disable Metrics/AbcSize
```
benlovell: Just disable it globally at the top of the file instead of repeating everywhere, especially since actually disabling it once in the file disables the rule for the rest of the file from that line.

benlovell: TIL but still pointless. It's a spec file so we apply different heuristics to production code. We should just globally disable this rule in the rubocop configuration:

```
# .rubocop.yml

Metrics/AbcSize:
  Exclude:
    - 'spec/features/*'
```

AndrewVos: @benlovell Fair enough. I would do that as another PR though, because of all the deletions that will have to happen.

PR #211, opened 2016-11-10
Display only my appointments

benlovell:
When the user is both a resource manager and a guider - the scoping logic
for the appointments JSON would short-circuit to all appointments due to
ordering of conditionals favouring the resource manager permission.

This change ensures that when the user is both guider and manager they only
see appointments scoped directly to them - i.e. appointments they are
holding with the customer.

### DIFF
--- a/app/assets/javascripts/calendars/guider-appointments.es6
+++ b/app/assets/javascripts/calendars/guider-appointments.es6
@@ -13,7 +13,7 @@
         nowIndicator: true,
         slotDuration: '00:30:00',
         eventTextColor: '#fff',
-        events: '/appointments'
+        events: '/appointments?mine'
       };
 
       super.start(el);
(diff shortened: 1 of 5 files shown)


AndrewVos:
I'm wondering if the `guider_and_resource_manager` thing is too much of an edge case that we should avoid. A user should really only have one role, but it's easier for testing if they don't. Maybe it would make sense to have some sort of dev toolbar that allows us to click `Be a guider` etc.

This would mean that we don't test with edge cases and also that we get to see the navbar how it would look for real.

benlovell:
@AndrewVos Good points. In practice we definitely will have users possessing multiple roles, particularly the guider / resource manager combination - so it's definitely something we need to guard against.

I'm sure in future it would make sense for us to add a roles multi-select box, similar to the organisations one I added to planner.

PR #213, opened 2016-11-11
Show valid breadcrumb for edit appointments depending on referer

AndrewVos:
I'm not sure I like the way this works. Thoughts @Guntrisoft @benlovell?

### DIFF
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -1,0 +1,29 @@
+module BreadcrumbHelper
+  def breadcrumb_part_for_previous_page
+    path, title = path_and_title_from_referer
+
+    content_tag(:li) do
+      content_tag(:a, href: path) do
+        title
+      end
+    end
+  end
+
+  private
+
+  def path_and_title_from_referer
+    if request_uri.start_with?(search_appointments_path)
+      [request_uri, 'Search']
+    elsif request.start_with?(activities_path)
+      [request_uri, 'My activity']
+    elsif request_uri.start_with?(calendar_path)
+      [request_uri, 'My appointments']
+    end
+  end
+
+  def request_uri
+    @request_uri ||= URI.parse(request.referer).request_uri
+  rescue URI::InvalidURIError
+    search_appointments_path
+  end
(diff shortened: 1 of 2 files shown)


benlovell:
You'll definitely need to be more defensive against the absence of a referrer. That seems to be fairly unreliable in practice.

AndrewVos:
@benlovell Yep, done

app/helpers/breadcrumb_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+  private
+
+  def path_and_title_from_referer
+    return request_uri, 'Search' if request_uri.start_with?(search_appointments_path)
+    return request_uri, 'My activity' if request_uri.start_with?(activities_path)
+    return request_uri, 'My appointments' if request_uri.start_with?(calendar_path)
+  end
+
+  def request_uri
+    URI.parse(request.referer).request_uri
```
benlovell: This could be memoized I think.

app/helpers/breadcrumb_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+      content_tag(:a, href: path) do
+        title
+      end
+    end
+  end
+
+  private
+
+  def path_and_title_from_referer
+    return request_uri, 'Search' if request_uri.start_with?(search_appointments_path)
```
benlovell: Stylistically i'm not a huge fan of these stacked returns, I wonder if building up a case statement with the paths interpolated into regex would be neater, also more defensive in that it could fall through to a catch-all, else case. This is nit-picky though and doesn't block you from merging.

app/helpers/breadcrumb_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+  end
+
+  private
+
+  def path_and_title_from_referer
+    if request_uri.start_with?(search_appointments_path)
+      [request_uri, 'Search']
+    elsif request.start_with?(activities_path)
```
Guntrisoft: @AndrewVos This line is causing an error on master for me (request_uri not request?)

PR #217, opened 2016-11-11
Show BookableSlots on guider appointment calendar and resource calendar

AndrewVos:

### DIFF
--- a/app/controllers/bookable_slots_controller.rb
+++ b/app/controllers/bookable_slots_controller.rb
@@ -1,7 +1,29 @@
 class BookableSlotsController < ApplicationController
+  def index
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+    slots = slots.for_guider(current_user) if scoped_to_me?
+    render json: slots
+  end
+
   def available
     render json: BookableSlot
-      .with_guider_count(Time.zone.now.to_date, 6.weeks.from_now.to_date)
+      .with_guider_count(Time.zone.today, 6.weeks.from_now.to_date)
       .to_json
+  end
+
+  def scoped_to_me?
+    params.key?(:mine)
+  end
+
+  private
+
+  def start_at
+    params[:start].to_date.beginning_of_day
+  end
+
+  def end_at
+    params[:end].to_date.beginning_of_day
   end
 end
--- a/app/serializers/bookable_slot_serializer.rb
+++ b/app/serializers/bookable_slot_serializer.rb
@@ -1,0 +1,7 @@
+class BookableSlotSerializer < ActiveModel::Serializer
+  attribute :start_at, key: :start
+  attribute :end_at, key: :end
+  attribute :resourceId do
+    object.guider_id
+  end
+end
--- a/app/views/calendars/show.html.erb
+++ b/app/views/calendars/show.html.erb
@@ -7,6 +7,7 @@
 
 <div
   id="GuiderAppointmentsCalendar"
+  class="t-calendar"
   data-module="guider-appointments-calendar"
   data-default-date="<%= Time.zone.now %>"
   data-guider-id="<%= current_user.id %>"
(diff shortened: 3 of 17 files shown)


app/serializers/bookable_slot_serializer.rb
```diff
@@ -0,0 +1,7 @@
+class BookableSlotSerializer < ActiveModel::Serializer
+  attribute :start_at, key: :start
+  attribute :end_at, key: :end
+  attribute :resourceId do
```
benlovell: Kinda strange to use this style mixed with the style for the other keys?

config/routes.rb
```diff
@@ -26,6 +26,7 @@
     delete '/', on: :collection, action: :destroy
     get 'merged', on: :collection
   end
+  resources :bookable_slots, only: :index
   get '/bookable_slots/available', to: 'bookable_slots#available'
```
benlovell: Not overly fussed but maybe this declaration could be nested within the `resources` one above now?

app/controllers/bookable_slots_controller.rb
```diff
@@ -1,7 +1,30 @@
 class BookableSlotsController < ApplicationController
+  def index
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+    slots = slots.where(guider: current_user) if scoped_to_me?
+    json = ActiveModelSerializers::SerializableResource.new(slots).to_json
+    render json: json
```
benlovell: I think you can just `render json: slots` here and the serializer is correctly inferred...

app/controllers/bookable_slots_controller.rb
```diff
@@ -1,7 +1,30 @@
 class BookableSlotsController < ApplicationController
+  def index
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+    slots = slots.where(guider: current_user) if scoped_to_me?
```
benlovell: I wonder if this would be nicer defined as a scope on `BookableSlot` instead? It's not a huge amount of code so not overly concerned.

AndrewVos: done

app/controllers/bookable_slots_controller.rb
AndrewVos: done

app/controllers/bookable_slots_controller.rb
```diff
@@ -1,7 +1,29 @@
...
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+    slots = slots.for_guider(current_user) if scoped_to_me?
+    render json: slots
+  end
+
   def available
     render json: BookableSlot
       .with_guider_count(Time.zone.now.to_date, 6.weeks.from_now.to_date)
```
dwhenry: This could be `Time.zone.today` which is slightly shorter...

app/views/calendars/show.html.erb
```diff
@@ -7,6 +7,7 @@
 
 <div
   id="GuiderAppointmentsCalendar"
+  class='t-calendar'
```
benbarnett: Should be `"` in HTML

spec/features/guider_views_appointments_spec.rb
```diff
@@ -2,14 +2,25 @@
 require 'rails_helper'
 
 RSpec.feature 'Guider views appointments' do
+  let(:today) do
+    BusinessDays.from_now(3)
+  end
```
dwhenry:
```
let(:today) { BusinessDays.from_now(3) }
```
and can we use a better name as today should always be the current day.

AndrewVos: It actually is `today` because we use `travel_to` later on.

app/views/calendars/show.html.erb
AndrewVos: done

app/serializers/bookable_slot_serializer.rb
benlovell: That's not what I was saying - why use the block style when the others are using the `key`? It's no big deal but thought I'd clarify.

AndrewVos: Oh right! Good point. I'll change this next time I'm around that file

PR #218, opened 2016-11-14
Add bugmuncher for feedback from UAT

benbarnett:
Only applies to development and staging environments.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -47,3 +47,18 @@
     });
 <% end %>
 </script>
+
+<% if Rails.env.development? || Rails.env.staging? %>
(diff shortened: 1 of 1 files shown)


AndrewVos:
LGTM

app/views/layouts/application.html.erb
```diff
@@ -47,3 +47,18 @@
     });
 <% end %>
 </script>
+
+<% if Rails.env.development? || Rails.env.staging? %>
+  <script type="text/javascript">
+    var bugmuncher_options = {
+      api_key: "fe678532af029a3c85a2"
```
benbarnett: Yeah that's what I was thinking tbh, let's see if it works for us for a few days?

PR #221, opened 2016-11-14
Allow filtering on the resource calendar

Guntrisoft:
screen shot 2016-11-14 at 14 11 19

### DIFF
--- a/app/assets/javascripts/calendars/guiders-appointments.es6
+++ b/app/assets/javascripts/calendars/guiders-appointments.es6
@@ -9,7 +9,13 @@
         defaultView: 'agendaDay',
         resourceLabelText: 'Guiders',
         header: {
-          right: 'agendaDay timelineDay today jumpToDate prev,next'
+          right: 'filter agendaDay timelineDay today jumpToDate prev,next'
+        },
+        customButtons: {
+          filter: {
+            text: 'Filter',
+            click: this.filterClick.bind(this)
+          }
         },
         buttonText: {
           agendaDay: 'Horizontal',
@@ -19,7 +25,6 @@
         nowIndicator: true,
         slotDuration: '00:30:00',
         eventTextColor: '#fff',
-        resources: '/guiders',
         eventSources: [
           {
             url: '/appointments'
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/calendars/guiders-appointments.es6
```diff
@@ -9,7 +9,13 @@
         defaultView: 'agendaDay',
         resourceLabelText: 'Guiders',
         header: {
-          right: 'agendaDay timelineDay today jumpToDate prev,next'
+          right: 'filter agendaDay timelineDay today jumpToDate prev,next'
+        },
+        customButtons: {
+          filter: {
+            text: 'Filter',
```
benbarnett: `glyphicon glyphicon-filter` worth it or no?

Could potentially use these for the horizontal/vertical ones if we wanted to go crazy

screen shot 2016-11-14 at 14 11 02

Guntrisoft: As discussed, we'll re-look at this separately from this PR.

app/assets/javascripts/calendars/guiders-appointments.es6
```diff
@@ -34,19 +39,80 @@
...
+    }
+
+    filterClick() {
+      this.filterPanel.css({
+        top: this.filterButton.offset().top + this.filterButton.height(),
+        left: this.filterButton.offset().left - (this.filterPanel.width() / 4)
+      });
+
+      this.showFilterPanel();
+      $('.select2-search__field').focus();
```
benbarnett: Ideally this would have a class name we control – is that possible?

app/assets/javascripts/calendars/guiders-appointments.es6
```diff
@@ -34,19 +39,80 @@
...
+
+      if (filteredResources.length === 0) {
+        return callback(resources);
+      }
+
+      callback(filteredResources);
+    }
+
+    filterResources(resource) {
+      return $.inArray(resource.id, $.map(this.filterList, (id) => { return parseInt(id); })) > -1;
```
AndrewVos: Bit hard to understand this, could you extract a variable for the map stuff?

`
let ids = $.map(this.filterList, (id) => { return parseInt(id); });
return $.inArray(resource.id, ids) > -1;
`

Or better yet, store `ints` in `filterList`? I feel like I'm missing something entirely here, or is that logical ^^

app/assets/javascripts/calendars/guiders-appointments.es6
benbarnett: Also it should probably be moved to within the `this.showFilterPanel();` function

app/assets/javascripts/calendars/guiders-appointments.es6
benbarnett: I think breaking this down might be a bit easier to read.

app/assets/javascripts/calendars/guiders-appointments.es6
Guntrisoft: Nope :(

app/assets/javascripts/calendars/guiders-appointments.es6
Guntrisoft: I've changed it so filterList is stored always as an array of ints, simplifying the filterResources method

app/assets/javascripts/calendars/guiders-appointments.es6
Guntrisoft: done

app/views/resource_calendars/show.html.erb
```diff
@@ -5,12 +5,26 @@
 
 <h1 class="sr-only">Allocations</h1>
 
+<div id="resource-calendar-filter" class="resource-calendar-filter hide">
+  <%= label_tag(:filter_users, 'Filter by user/group') %>
```
benlovell: We should refer to guiders instead of users.

Guntrisoft: fixed

PR #224, opened 2016-11-14
Don't search six weeks of bookable slots for appointment slot picker

AndrewVos:
This speeds up slot rendering massively


Guntrisoft:
👍

PR #228, opened 2016-11-15
Calendar colours

Guntrisoft:
screen shot 2016-11-15 at 10 58 15

- attempted to make appointments stand out and make the calendar
  not so overwhelming with colour
- cancelled appointments now have the same hue as default appointment
  colour - so to not stand out too much
- holidays are now grey, so not to confuse with cancelled appointments

### DIFF
--- a/app/assets/javascripts/calendars/guiders-appointments.es6
+++ b/app/assets/javascripts/calendars/guiders-appointments.es6
@@ -32,12 +32,12 @@
           },
           {
             url: '/holidays',
-            color: 'red',
+            color: '#ebebeb',
             rendering: 'background'
           },
           {
             url: '/bookable_slots',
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/calendars/guiders-appointments.es6
```diff
@@ -32,12 +32,12 @@
           },
           {
             url: '/holidays',
-            color: 'red',
+            color: '#ebebeb',
```
Guntrisoft: Need to work on something to get these from the sass instead if possible

AndrewVos:
I LOVE that you chose grey. It's a lot less aggressive than the red/pink.

benlovell:
💓

PR #232, opened 2016-11-15
Separate concerns of the company and resource calendars

Guntrisoft:
- all the methods to do with editing appointments is now
  in the allocationsCalendar
- allocationCalendar extends from company calendar
- company calendar provides the basic functionality of a read
  only calendar for the guiders to view the whole company
- fixes issue with filter on company calendar not working

### DIFF
--- a/app/views/shared/_calendar_filter.html.erb
+++ b/app/views/shared/_calendar_filter.html.erb
@@ -1,0 +1,12 @@
+<div id="resource-calendar-filter" class="resource-calendar-filter hide">
+  <%= label_tag(:filter_users, 'Filter by guider/group') %>
+  <%=
+    select_tag(
+      :filter_guiders,
+      grouped_options_for_select(guider_options),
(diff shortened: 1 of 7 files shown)


app/views/shared/_calendar_filter.html.erb
```diff
@@ -0,0 +1,12 @@
+<div id="resource-calendar-filter" class="resource-calendar-filter hide">
+  <%= label_tag(:filter_users, 'Filter by guider/group') %>
+  <%=
+    select_tag(
+      :filter_guiders,
+      grouped_options_for_select(guider_options),
+      multiple: 'multiple',
+      class: 't-users',
+      data: { module: 'advanced-select', placeholder: 'Guider or Group' }
```
Guntrisoft: Had to add a placeholder as there is a bug in select2 which causes a JS error when using allowClear: true and having no placeholder

https://github.com/select2/select2/issues/3772

PR #233, opened 2016-11-15
Ensure there's a corresponding time being dragged to

benbarnett:
We were trying to calculate position of elements that didn't exist.
E.g. dragging an event to before 8.30am or after 7pm.

Also adjusts the 'boot' order to ensure `this.$el` is defined
before we add the row highlighter.


Guntrisoft:
👍

PR #234, opened 2016-11-15
Make resource calendar the homepage for resource managers

benbarnett:
Seems to be a more frequently used page than the groups management.


Guntrisoft:
👍

benlovell:
Cool. As an aside we should move these to the routing and cut out a bunch of the stack (at some point)

PR #235, opened 2016-11-15
Alter the colours of some calendar events

Guntrisoft:
- holidays calendar now has grey events, in tune
  with the grey background events on resource calendar
- bookable slots and holidays on resource calendar and
  guider appointments calendar now has colour from
  class name rather than colour hard coded into JS


benbarnett:
Could you whack a screenshot on?

PR #236, opened 2016-11-15
Make schedule edit screen less painful on the eyes

benbarnett:
Not sure about this colour but it sets up its own variable
for easy tweaking in the future.

Before:
screen shot 2016-11-15 at 16 03 07

Now:
screen shot 2016-11-15 at 16 18 56


Guntrisoft:
👍

PR #238, opened 2016-11-16
Resource managers can sort guiders

AndrewVos:
# New button:

# The sort page:

### DIFF
--- a/app/assets/javascripts/calendars/company.es6
+++ b/app/assets/javascripts/calendars/company.es6
@@ -6,12 +6,16 @@ class CompanyCalendar extends Calendar {
       defaultView: 'agendaDay',
       resourceLabelText: 'Guiders',
       header: {
-        right: 'filter agendaDay timelineDay today jumpToDate prev,next'
+        right: 'sort filter agendaDay timelineDay today jumpToDate prev,next'
       },
       customButtons: {
         filter: {
           text: 'Filter',
           click: this.filterClick.bind(this)
+        },
+        sort: {
+          text: 'Sort',
+          click: this.sortClick.bind(this)
         }
       },
       buttonText: {
@@ -112,6 +116,10 @@ class CompanyCalendar extends Calendar {
     });
 
     this.showFilterPanel();
+  }
+
+  sortClick() {
+    window.location = '/users/sort';
   }
 
   eventAfterRender(event, element) {
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -14,4 +14,20 @@ class UsersController < ApplicationController
         .by_start_at
     )
   end
+
+  def sort
+    @guiders = User.guiders.includes(:groups)
+  end
+
+  def sort_update
+    guider_order = params[:order_users][:guider_order]
+
+    ActiveRecord::Base.transaction do
+      guider_order.each_with_index do |id, index|
+        User.update(id, position: index)
+      end
+    end
+
+    redirect_to :sort_users, success: 'Guider order has been updated'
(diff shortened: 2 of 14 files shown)


app/controllers/users_controller.rb
```diff
@@ -14,4 +14,20 @@ def edit
         .by_start_at
     )
   end
+
+  def sort
+    @guiders = User
```
benlovell: May as well just collapse this on one line, it won't be particularly long.

app/controllers/users_controller.rb
```diff
@@ -14,4 +14,20 @@ def edit
...
+               .guiders
+               .includes(:groups)
+  end
+
+  def sort_update
+    guider_order = params[:order_users][:guider_order]
+    ActiveRecord::Base.transaction do
```
benlovell: Can you put some breathing space around this? Just a line of whitespace before and after should make this somewhat easier to scan.

app/assets/javascripts/calendars/company.es6
```diff
@@ -6,12 +6,16 @@ class CompanyCalendar extends Calendar {
       defaultView: 'agendaDay',
       resourceLabelText: 'Guiders',
       header: {
-        right: 'filter agendaDay timelineDay today jumpToDate prev,next'
+        right: 'sort filter agendaDay timelineDay today jumpToDate prev,next'
```
benbarnett: We could probably just make this a top-level nav thing. But I'm not sure where it would go.

Hmm.

db/migrate/20161116103731_add_position_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :position, :integer
```
benlovell: Can we make this not nullable and default it to `0`?

app/assets/javascripts/calendars/company.es6
AndrewVos: Yeah totally

spec/features/resource_manager_sorts_guiders_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def when_they_sort_those_guiders
+    @page = Pages::SortGuiders.new.tap(&:load)
+    @page.reverse_guiders
+    @page.save.click
+  end
+
+  def then_the_guider_order_is_preserved
+    expect(User.guiders).to eq @guiders.reverse
```
benlovell: We should make this same assertion but based on the ordering of the guiders in the UI instead. Also perhaps saying the order was preserved is incorrect, given they've all just been changed?

app/assets/javascripts/calendars/company.es6
benbarnett: Yeah that's actually a really good idea

app/controllers/users_controller.rb
AndrewVos: Yup

app/controllers/users_controller.rb
AndrewVos: k

db/migrate/20161116103731_add_position_to_users.rb
AndrewVos: k

app/assets/javascripts/calendars/company.es6
benbarnett: Probably best to separate out as we're renaming the thing too

PR #239, opened 2016-11-16
Fix flaky sull text search test

AndrewVos:
We were getting failures because the appointments factory generates
random names and sometimes either the first name or the last name was
not distinct.


AndrewVos:
:smile:

benlovell:
Sull text. Nice work.

PR #241, opened 2016-11-16
Add 'list' view to guider's appointments page

benbarnett:
Makes it easier to view the day at a glance.

Bit like this:

screen shot 2016-11-16 at 13 39 21


Guntrisoft:
👍

PR #242, opened 2016-11-16
Ensure holidays appear over the top of bookable slots

benbarnett:
When bookable slots starting after a holiday's start time, they would appear over the top of the holiday. This resulted in the guider appearing to be available when they were in fact not.

**See example below for a holiday that ends at 14:00, but appears to end earlier**

Before:
screen shot 2016-11-16 at 14 28 05

After:
screen shot 2016-11-16 at 14 27 39


Guntrisoft:
👍

PR #243, opened 2016-11-16
Indicate filters are applied by showing on button label

benbarnett:
Show the number of filters applied so resource managers know
that there is one applied to the current view.

i9fy31a7oe


Guntrisoft:
👍 nice

PR #244, opened 2016-11-16
Users can rebook appointments

AndrewVos:
# New button:

# Details are copied over:

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -44,7 +44,7 @@ class AppointmentsController < ApplicationController
   end
 
   def new
-    @appointment = Appointment.new
+    @appointment = copy_appointment || Appointment.new
   end
 
   def create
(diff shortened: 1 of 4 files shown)


app/controllers/appointments_controller.rb
```diff
@@ -131,4 +131,12 @@ def redirect_after_successful_update
       redirect_to calendar_path, success: 'Appointment has been modified'
     end
   end
+
+  def copy_appointment
+    return nil unless params[:copy_from].present?
+    appointment = Appointment.find(params[:copy_from]).dup
+    appointment.start_at = nil
+    appointment.end_at = nil
+    appointment
```
benbarnett: Is there any other stuff like this? Associated guider maybe? Activity?

AndrewVos: The other stuff isn't rendered as inputs in the view, so they can't get saved. The only two afaik are those, cause they're actually hidden fields in the view.

benbarnett: Ah yeah, ok cool

benbarnett:
😍

PR #245, opened 2016-11-16
Better flash messages

benbarnett:
Also handles failure cases. Alerts automatically fade out after 3 seconds.

screen shot 2016-11-16 at 15 45 08
screen shot 2016-11-16 at 15 45 24


Guntrisoft:
👍 top work

PR #246, opened 2016-11-16
[DO NOT MERGE] Stop the user from losing changes to a schedule

Guntrisoft:
- if the user makes any changes and doesn't hit the save
  button then they will be alerted that they have changes
  that will be lost

### DIFF
--- a/app/assets/javascripts/calendars/guider-slot-picker.es6
+++ b/app/assets/javascripts/calendars/guider-slot-picker.es6
@@ -20,13 +20,72 @@
 
       super.start(el);
 
-      this.addEvents();
-      this.generateJSON();
-      this.setupEvents();
+      this.saveWarningMessage = 'You have unsaved changes - Save, or undo the changes.';
+      this.hiddenDataInput = $(this.$el.data('events'));
+      this.saveButton = $('.js-button-save');
+
+      this.addExistingEventsToCalendar();
+
+      this.eventsAtPageLoad = this.getJSON();
+
+      this.saveData();
+      this.bindEvents();
+    }
+
+    addExistingEventsToCalendar() {
+      const events = JSON.parse($(this.$el.data('events')).val()),
+      calendarView = this.$el.fullCalendar('getView'),
+      calendarStartDate = calendarView.intervalStart,
+      calendarEndDate = calendarView.intervalEnd.subtract(2, 'days'),
+      eventsToAdd = [];
+
+      for (let currentDate = moment(calendarStartDate); currentDate < calendarEndDate; currentDate.add(1, 'days')) {
+        for (let eventIndex in events) {
+          let event = events[eventIndex];
+          if (event.day_of_week == currentDate.day()) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -20,13 +20,72 @@
 
       super.start(el);
 
-      this.addEvents();
-      this.generateJSON();
-      this.setupEvents();
+      this.saveWarningMessage = 'You have unsaved changes - Save, or undo the changes.';
```
benbarnett: Should this be in `TapBase` or `Calendar` now?

app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -20,13 +20,72 @@
...
+      this.hiddenDataInput = $(this.$el.data('events'));
+      this.saveButton = $('.js-button-save');
+
+      this.addExistingEventsToCalendar();
```
benbarnett: This constructor is getting a little messy now 

Can we move these couple of lines into an `initCalendar()` method or something?

app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -105,11 +151,27 @@
...
       }
 
-      dataElement.val(JSON.stringify(eventsOutput, null, 2));
+      return JSON.stringify(eventsOutput.sort(this.sortEvents.bind(this)));
     }
 
-    setupEvents() {
-      $(`#${this.$el.data('events-common')}`).find('button').on('click', this.handleEvent.bind(this));
+    calculateSortFingerprint(event) {
+      const minutes = (event.start_hour * 60) + event.start_minute;
```
benbarnett: 🙀

Do we have access to `moment` here?

app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -122,20 +184,23 @@
       const events = $(event.currentTarget).data('events'),
       calendarView = this.$el.fullCalendar('getView'),
       calendarStartDate = calendarView.intervalStart,
-      calendarEndDate = calendarView.intervalEnd;
+      calendarEndDate = calendarView.intervalEnd.subtract(2, 'days'),
+      eventsToAdd = [];
 
       for (let currentDate = moment(calendarStartDate); currentDate < calendarEndDate; currentDate.add(1, 'days')) {
         for (let eventIndex in events) {
           event = events[eventIndex];
```
benbarnett: Have I misread this or are we overwriting the argument to the function here? We might have got round a missing lint warning here.

app/assets/javascripts/calendars/guider-slot-picker.es6
Guntrisoft: It's not strictly the constructor anymore (it's the start method), but I get your jist 👍 - will split out.

PR #253, opened 2016-11-17
Only use the floating 'fixed' flash message style on allocations

benbarnett:
As we use AJAX-style saving on this calendar, it makes sense to
flash the message and then fade it out. But on other pages
this just looks a little strange, so they get the standard
full width, block style message.

This essentially fixes the following:
screen shot 2016-11-17 at 11 10 02

with:
screen shot 2016-11-17 at 11 10 08


Guntrisoft:
👍

PR #254, opened 2016-11-17
Balance column widths on guider list page

benbarnett:
Seems the groups are carrying the weight of the content on this
page, so this gives them a little more space.

Before:
screen shot 2016-11-17 at 11 34 38

After:
screen shot 2016-11-17 at 11 34 53


Guntrisoft:
👍

PR #255, opened 2016-11-17
Stop the user from losing changes to a schedule

Guntrisoft:
- If the user makes any changes and doesn't hit the save
button then they will be alerted that they have changes
that will be lost

### DIFF
--- a/app/assets/javascripts/calendars/guider-slot-picker.es6
+++ b/app/assets/javascripts/calendars/guider-slot-picker.es6
@@ -20,13 +20,34 @@
 
       super.start(el);
 
+      this.initialEventTimestamps = [];
+
       this.addEvents();
       this.generateJSON();
-      this.setupEvents();
+      this.bindEvents();
+    }
+
+    bindEvents() {
+      $(`#${this.$el.data('events-common')}`).find('button').on('click', this.handleEvent.bind(this));
     }
 
     eventDrop() {
+      this.clearUnloadEvent();
       this.generateJSON();
+    }
+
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -20,22 +20,44 @@
...
+    }
+
     select(start) {
       const event = {
         start: start,
         end: moment(start).add(this.config.slotDurationMinutes, 'minutes')
       };
 
       if (!this.isOverlapping(event)) {
+        this.clearUnloadEvent();
```
AndrewVos: Why is this here?

app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -44,9 +66,14 @@
     }
 
     eventRender(event, element) {
+      if ($.inArray(event.start.unix(), this.initialEventTimestamps) === -1) {
+        this.setUnloadEvent();
+      }
+
```
AndrewVos: Oh wow, this is beautiful!

app/assets/javascripts/calendars/guider-slot-picker.es6
Guntrisoft: We need to reset the unload event, any time something happens to the calendar

AndrewVos: Oh right I see

PR #256, opened 2016-11-17
Configure mailers

benlovell:
Adds mailgun specific headers for identification.

Changes the `from` address so we're unique from planner and other apps we
send email from.


AndrewVos:
LG, thanks!

PR #257, opened 2016-11-17
Ensure our calendar styles apply to month views

benbarnett:
We need to be quite specific to override FullCalendar's styles.

Before:
screen shot 2016-11-17 at 13 31 07

After:
screen shot 2016-11-17 at 13 31 14


Guntrisoft:
👍

PR #258, opened 2016-11-17
Show holidays in guider calendar

AndrewVos:


app/models/holiday.rb
```diff
@@ -18,6 +18,10 @@ def self.merged_for_calendar_view
       .order(:start_at)
   end
 
+  def self.scoped_to_user(user)
+    where(user_id: [user.id, nil])
```
benlovell: What's the array about?

benlovell: OK how about something like `where(user_id: [user.id, BANK_HOLIDAYS = nil])` so it's a little more descriptive?

AndrewVos: I don't really like that it's a bit confusing at first and rubocop would complain about the dynamic assignment. I changed the method name and rewrote it to be more obvious.

PR #260, opened 2016-11-17
Fixes issue with slot picker and the unload event

Guntrisoft:
- when you remove a slot and there are no events to render
  the check on whether to set the unload event no longer happens.
  This caters for that scenario


app/assets/javascripts/calendars/guider-slot-picker.es6
```diff
@@ -76,11 +76,22 @@
...
+    }
+
+    handleCloseEvent() {
+      this.clearUnloadEvent();
+
+      this.$el.fullCalendar('removeEvents', event._id);
+
+      if (
+        this.$el.fullCalendar('clientEvents').length === 0 &&
+        this.initialEventTimestamps.length > 0
```
AndrewVos: can you do != instead?

PR #265, opened 2016-11-18
When events overlap, make their widths more even

benbarnett:
As we're dealing with quite small columns, the events display
a little odd.

More information here: https://fullcalendar.io/docs/agenda/slotEventOverlap/


Guntrisoft:
👍

PR #269, opened 2016-11-18
Appointment status changes/addition

Guntrisoft:
- Altered Completed to Complete
- Added Incomplete

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -6,8 +6,9 @@ class Appointment < ApplicationRecord
 
   enum status: %i(
     pending
-    completed
+    complete
     no_show
+    incomplete
     ineligible_age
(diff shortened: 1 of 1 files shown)


Guntrisoft:
This will have an affect on existing appointments on staging, as I'm shifting the enum index by adding in incomplete - @benbarnett suggests this should be ok as it's all test data at the moment

benlovell:
Cool. No need to summarise the diff in the commit message though. Just some text to describe _why_ you're making the change is preferable.

benbarnett:
@benlovell @Guntrisoft agree with the commit message stuff. I've tried to put the 'why' in the Trello ticket description on each of these, so maybe just use that next time 👍🏻

PR #270, opened 2016-11-18
Speed up sass compilation with native extensions

benlovell:


benlovell:
@benbarnett Seemed OK given my cursory glance.

PR #272, opened 2016-11-18
Users can search for appointments by guider name

AndrewVos:
screenshot from 2016-11-18 14 06 22

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -1,7 +1,8 @@
 class Appointment < ApplicationRecord
   include PgSearch
 
-  pg_search_scope :search, against: %i(id first_name last_name)
+  pg_search_scope :search, against: %i(id first_name last_name), associated_against: { guider: :name }
+
   belongs_to :agent, class_name: 'User'
 
   enum status: %i(
--- a/spec/models/appointment_spec.rb
+++ b/spec/models/appointment_spec.rb
@@ -193,6 +193,15 @@ RSpec.describe Appointment, type: :model do
       results = results(nil, date_range_start, date_range_end)
       expect(results).to eq [appointment]
     end
+
+    it 'returns results for guider name' do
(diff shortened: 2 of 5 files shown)


spec/models/appointment_spec.rb
```diff
@@ -193,6 +193,15 @@ def results(q, start_at, end_at)
       results = results(nil, date_range_start, date_range_end)
       expect(results).to eq [appointment]
     end
+
+    it 'returns results for guider name' do
+      guider = create(:guider, name: 'Kate Bush')
```
AndrewVos: :ghost: :door:

AndrewVos: no window emoji

app/models/appointment.rb
```diff
@@ -1,7 +1,10 @@
 class Appointment < ApplicationRecord
   include PgSearch
 
-  pg_search_scope :search, against: %i(id first_name last_name)
+  pg_search_scope :search, against: %i(id first_name last_name), associated_against: {
+    guider: [:name]
```
benlovell: How does this search work? Is it an exact match and case-sensitive?

AndrewVos: They do, but I thought leaving the array form in there would show the next person in this file that they can just add another key if they want, instead of having to read the docs.

Maybe that's dumb though. I'll change it.

benbarnett: Actually they brought up the search thing yesterday, ideally we would have partial matching on this stuff. But I thought I'd bring that up later in case it's a big change 🙈

AndrewVos: @benlovell it's not case sensitive, and afaik does porter stemming. Some more specs might be warranted.

It does *not* do partial matches, and actually after some testing I haven't seen it do porter stemming either.

PR #274, opened 2016-11-18
Clean up application.js

Guntrisoft:
- moved JS modules into modules directory as per govuk admin
  standard
- removed lines from application.js which don't need to be there
  as 'require_tree .' will import them


AndrewVos:
<3

PR #280, opened 2016-11-21
Fix IE11 issue with unload event on slot picker edit page

Guntrisoft:
- saving unix timestamps was causing an issue in IE11 where
  date matching was an hour out (timezone issue possibly?)
- replaced unix timestamps in favour of date strings
- fixed another issue where IE11 was alerting multiple times

### DIFF
--- a/app/assets/javascripts/modules/calendars/guider-slot-picker.es6
+++ b/app/assets/javascripts/modules/calendars/guider-slot-picker.es6
@@ -20,7 +20,8 @@
 
       super.start(el);
 
-      this.initialEventTimestamps = [];
+      this.initialEventDateTimes = [];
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/calendars/guider-slot-picker.es6
```diff
@@ -44,14 +54,10 @@
       $(window).on('beforeunload', () => {
         return this.saveWarningMessage;
       });
-
-      $(window).on('unload', () => {
```
Guntrisoft: As far as I can see, Chrome/Firefox/IE11 are all happy with just `beforeunload` event and not `unload`

PR #281, opened 2016-11-21
Validate invalid date of births

AndrewVos:
For example 31st February was being converted to 3rd May.

image

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -31,6 +31,7 @@ class Appointment < ApplicationRecord
 
   validate :not_within_two_business_days
   validate :not_more_than_thirty_business_days_in_future
+  validate :date_of_birth_valid
 
   has_many :activities, -> { order('created_at DESC') }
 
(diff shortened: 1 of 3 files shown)


app/models/appointment.rb
```diff
@@ -31,11 +31,26 @@ class Appointment < ApplicationRecord
 
   validate :not_within_two_business_days
   validate :not_more_than_thirty_business_days_in_future
+  validate :date_of_birth_valid
 
   has_many :activities, -> { order('created_at DESC') }
 
   audited on: %i(create update)
 
+  def date_of_birth=(value)
```
benlovell: Move this down with the other instance methods for consistency.

app/models/appointment.rb
```diff
@@ -31,11 +31,26 @@ class Appointment < ApplicationRecord
...
+    if value.is_a?(Hash)
+      super(Date.new(value[1], value[2], value[3]))
+    else
+      super(value)
+    end
+  rescue ArgumentError
+    @date_of_birth_invalid = true
+  end
+
+  def date_of_birth_valid
```
benlovell: Move this down under the `private` section.

AndrewVos: done

PR #283, opened 2016-11-21
Associate rebooked appointments with original

benlovell:
Stores the original appointment when rebooking an appointment from a
previous one.

## New

screen shot 2016-11-21 at 11 55 31

## Editing

screen shot 2016-11-21 at 11 29 08


app/helpers/appointment_helper.rb
```diff
@@ -2,4 +2,24 @@ module AppointmentHelper
   def friendly_options(statuses)
     statuses.map { |k, _| [k.titleize, k] }.to_h
   end
+
+  def rebooked_from(appointment)
+    return unless appointment.rebooked_from_id?
+
+    content_tag(
+      :span,
+      "rebooked from ##{appointment.rebooked_from_id}",
```
benbarnett: Not a big deal but would be nice for this to be a link back to this appointment maybe?

benlovell: I slipped a change in quick so maybe this is moot now but I didn't think it made sense linking back as I couldn't imagine a flow where they'd want to go back to the original appointment they're booking from, especially since the original's details are copied here anyway.

Happy to alter though if anyone feels strongly about it.

benlovell: That said - I guess it does make more sense in the edit scenario at least...

benbarnett: I guess one scenario is quickly looking back to see when this last appointment was. 

This is making me think we could do with a 'quick jump' thing to go straight to an appointment with a given reference. Maybe we just think about this and tackle separately?

PR #284, opened 2016-11-21
Accessibility fixes

Guntrisoft:
- Ensuring form elements have labels
- Ensuring there is enough colour contrast on the breadcrumb
- Just hide the contents of the events on the resource calendar
  from visual users, rather than removing it
- Ensuring that icon buttons on full calendar have
  hidden screen reader text alternatives

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,3 +1,4 @@
 @import "lib/*";
-@import "vendor/vendor";
+@import "vendor/*";
 @import "components/*";
+@import "helper/*";
--- a/app/views/shared/_date_of_birth_form_field.html.erb
+++ b/app/views/shared/_date_of_birth_form_field.html.erb
@@ -1,13 +1,18 @@
-<% start_year = 120.years.ago %>
+<%
+  start_year = 120.years.ago
+  form_element_day_id = "#{form.object_name}_date_of_birth_3i"
+  form_element_month_id = "#{form.object_name}_date_of_birth_2i"
+  form_element_year_id = "#{form.object_name}_date_of_birth_1i"
+%>
 
 <div class="form-dob clearfix" data-module="customer-age" data-output-id="js-customer-age">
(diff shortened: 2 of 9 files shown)


app/assets/stylesheets/lib/_bootstrap_overrides.scss
```diff
@@ -0,0 +1,7 @@
+.breadcrumb > .active {
```
benbarnett: Previously we only used `lib` for SASS functions/variables/mixins that didn't actually output any CSS. Perhaps this should be a `helper` or something? 

This would also clear up the load order weirdness.

app/views/shared/_date_of_birth_form_field.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="form-dob clearfix" data-module="customer-age" data-output-id="js-customer-age">
   <%= field_with_errors_wrapper(form, :date_of_birth, 'form-dob__inputs-container') do %>
-    <label>
+    <label for="<%= form.object_name %>_date_of_birth_3i">
```
benbarnett: Maybe we should extract these as variables?

app/assets/stylesheets/application.scss
```diff
@@ -1,3 +1,4 @@
 @import "lib/*";
 @import "vendor/vendor";
 @import "components/*";
+@import "helper/bootstrap_overrides";
```
benbarnett: For consistency, we should probably either use globbing or not use globbing if you know what I mean

app/assets/stylesheets/lib/_bootstrap_overrides.scss
Guntrisoft: sorted

app/views/shared/_date_of_birth_form_field.html.erb
Guntrisoft: done

app/assets/stylesheets/application.scss
Guntrisoft: agreed - sorted

PR #285, opened 2016-11-21
Rerender edit page after editing an appointment

AndrewVos:
Also, change the way that we store the previous page so that we can
always show the correct breadcrumb on appointments#edit.

### DIFF
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -1,33 +1,38 @@
 module BreadcrumbHelper
   def breadcrumb_part_for_previous_page
-    path, title = path_and_title_from_referer
-
     content_tag(:li) do
-      content_tag(:a, href: path) do
-        title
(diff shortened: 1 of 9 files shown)


app/helpers/breadcrumb_helper.rb
```diff
@@ -11,23 +11,17 @@ def breadcrumb_part_for_previous_page
...
-    end
-  end
+  def previous_page_info
+    page = session[:previous_page].to_s
 
-  def request_uri
-    @request_uri ||= URI.parse(request.referer).request_uri
-  rescue URI::InvalidURIError
-    search_appointments_path
+    if page == 'activities#index'
```
benlovell: Can you switch this to a `case` statement instead?

AndrewVos: I switched to using a hash instead. Likey?

benlovell:
👍

PR #287, opened 2016-11-21
Only link activity feed items from my activity page

Guntrisoft:
On the edit page, each activity was a link back to the same page - this is not great for screen reader users, when they view the list of available links on the page.

### DIFF
--- a/app/views/activities/_activity.html.erb
+++ b/app/views/activities/_activity.html.erb
@@ -1,7 +1,15 @@
-<a class="list-group-item t-activity" href="<%= edit_appointment_path(activity.appointment) %>">
-  <span class="glyphicon glyphicon-calendar" aria-hidden="true"></span>
-  <b><%= activity.user_name %></b>
-  <%= content_for(:activity_message) %>
-  <%= render('activities/details', appointment: activity.appointment) if details %>
-  <em class="badge"><%= activity.created_at.to_s(:govuk_date_short) %> (<%= time_ago_in_words(activity.created_at)%> ago)</em>
-</a>
+<% if details %>
+  <a class="list-group-item t-activity" href="<%= edit_appointment_path(activity.appointment) %>">
+<% else %>
+  <li class="list-group-item t-activity">
+<% end %>
+    <span class="glyphicon glyphicon-calendar" aria-hidden="true"></span>
+    <b><%= activity.user_name %></b>
+    <%= content_for(:activity_message) %>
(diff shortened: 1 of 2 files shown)


benlovell:
Is this really a problem?

Guntrisoft:
1) the markup was invalid in it's current state (we have had `UL` > `A`)
2) this was creating unnecessary links back to the same page on the edit appointment page - which is bad for accessibility/screen readers

Guntrisoft:
I need to do some more work in order to get this right - the context in where the activity is used determines whether to make it a link or not.

Guntrisoft:
Ok this should be good now!

app/jobs/pusher_activity_notification_job.rb
```diff
@@ -7,21 +7,21 @@ def perform(recipient, activity)
     Pusher.trigger(
       'telephone_appointment_planner',
       "guider_activity_#{recipient.id}",
-      body: render_activity(activity, true)
+      body: render_activity(activity, true, true)
```
AndrewVos: Passing booleans into methods like this is generally not a good idea, unless the method name makes it obvious what the boolean is for. Something like this would be more readable:

```
render_activity(details: false)

def render_activitiy(activity, options={})
  options = {details: true}.merge(options)
end
```

And now you can have details default to true, so that you only need to pass in options when you want.

app/views/activities/_activity.html.erb
```diff
@@ -1,7 +1,15 @@
...
-  <%= content_for(:activity_message) %>
-  <%= render('activities/details', appointment: activity.appointment) if details %>
-  <em class="badge"><%= activity.created_at.to_s(:govuk_date_short) %> (<%= time_ago_in_words(activity.created_at)%> ago)</em>
-</a>
+<% if link %>
+  <a class="list-group-item t-activity" href="<%= edit_appointment_path(activity.appointment) %>">
+<% else %>
+  <li class="list-group-item t-activity">
+<% end %>
+<span class="glyphicon glyphicon-calendar" aria-hidden="true"></span>
```
benbarnett: Can we indent these lines as if they're in the tag? Might make it a bit more readable.

Guntrisoft: done

benbarnett:
Can you add the PR description to the commit itself too?

Guntrisoft:
@benbarnett done

PR #288, opened 2016-11-21
remove attribute 'pattern' which is not required on a number field

Guntrisoft:
type=number automatically means that the pattern is [0-9]* - so no need to have it as an attribute.


Guntrisoft:
from HTML5 validator: `Attribute pattern is only allowed when the input type is email, password, search, tel, text, or url`

benbarnett:
Can we invert this change? Add `pattern` back in but use `type="text"`?

benlovell:
I'll pick this up separately now.

PR #290, opened 2016-11-22
Better validation for date of birth components

benlovell:
Makes the date of birth components required (client-side) plus:

Checks:

- the year is numeric, starts with 19 and is 4 digits
- the month is numeric, between 01-12 and max 2 digits
- the day is numeric, between 01-31 and max 2 digits

The only catch is that I'm not certain this kicks up the numeric keyboard on mobile devices.

### DIFF
--- a/app/views/shared/_date_of_birth_form_field.html.erb
+++ b/app/views/shared/_date_of_birth_form_field.html.erb
@@ -19,6 +19,7 @@
           pattern: '(0?[1-9]|(1|2)[0-9]|3[0-1])',
           maxlength: 2,
           required: true,
+          title: 'Must be numeric eg 01',
           'aria-describedby': 'dob-hint'
         )
       %>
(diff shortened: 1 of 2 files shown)


app/views/shared/_date_of_birth_form_field.html.erb
```diff
@@ -10,49 +9,49 @@
...
           value: form.object.date_of_birth.try(:day),
           placeholder: 'DD',
           class: 'form-dob__input js-dob-day t-date-of-birth-day',
-          pattern: '[0-9]*',
-          min: 1,
-          max: 31,
+          pattern: '(0?[1-9]|(1|2)[0-9]|3[0-1])',
+          maxlength: 2,
+          required: true,
           'aria-describedby': 'dob-hint'
```
benbarnett: If we add a `title` attribute to these fields, it'll be displayed as part of the error.

So perhaps `title="Must be numeric"` or something?

screen shot 2016-11-22 at 14 00 20

benlovell: Ah awesome. Will do.

PR #295, opened 2016-11-22
Drag an area on the allocations calendar to add a holiday

AndrewVos:
image

### DIFF
--- a/app/assets/javascripts/modules/calendars/allocations.es6
+++ b/app/assets/javascripts/modules/calendars/allocations.es6
@@ -25,6 +25,9 @@
       this.actionPanel = $('[data-action-panel]');
       this.$savedChanges = $('.js-saved-changes');
       this.$form = $('.js-changes-form');
+      this.$holidayForm = $('.js-holiday-form');
+      $(this.$holidayForm).on('ajax:success', this.afterHolidayCreateSuccess.bind(this));
+      $(this.$holidayForm).on('ajax:error', this.afterHolidayCreateError.bind(this));
 
       super.start(el);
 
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/modules/calendars/allocations.es6
```diff
@@ -38,19 +41,11 @@
...
-        console.log('Holiday guider name', resource.title);
-
-        // @todo Jquery ajax request to POST a new holiday
-        // in the onsuccess method we will need to call
-        // this.$el.fullCalendar('refetchEvents');
-        // this.showAlert('.alert-success');
-        //
-        // or call this.showAlert('.alert-danger');
-        // on failure
+        this.$holidayForm.find('.js-user-id').val(resource.id);
```
Guntrisoft: Any reason why this is not done as a `$.post()` https://api.jquery.com/jquery.post/ ?

AndrewVos: Mostly cause I started to get all paranoid about CSRF and thought I should do the safest thing.
I'm not 100% sure that's valid I should probably learn rails a bit more 🤑

app/assets/javascripts/modules/calendars/allocations.es6
```diff
@@ -16,14 +16,18 @@
...
+        },
+        selectable: true
       };
 
       this.eventChanges = [];
       this.isFullscreen = false;
       this.actionPanel = $('[data-action-panel]');
       this.$savedChanges = $('.js-saved-changes');
       this.$form = $('.js-changes-form');
+      this.$holidayForm = $('.js-holiday-form');
```
AndrewVos: Yeah I'm all for that jquery stuff. I hated it in the beginning thinking it was like prefixing the type (`strName`) but it's really really good for readibility!

Guntrisoft: yup, totally agree! - would be great if we can enforce it with a linting rule?

app/assets/javascripts/modules/calendars/allocations.es6
```diff
@@ -33,6 +37,20 @@
       this.setupUndo();
     }
 
+    select(start, end, jsEvent, view, resource) {
+      let title;
+
+      if (title = prompt(`Name of holiday period for ${resource.title}?`)) {
```
AndrewVos: @Guntrisoft I can inline the title variable here right?

benlovell:
As discussed, I think `batch_create` makes sense over `create_batch`.

PR #296, opened 2016-11-23
Fix bug where customer age module was throwing warnings

Guntrisoft:
**BEFORE**

screen shot 2016-11-23 at 10 19 18

**AFTER**

screen shot 2016-11-23 at 10 19 50

- If you entered letters into the text fields for DOB
  it resulted in console warning errors from moment.js
- Beefed up the checking of the values for DOB to ensure
  this doesn't happen

### DIFF
--- a/app/assets/javascripts/modules/customer-age.es6
+++ b/app/assets/javascripts/modules/customer-age.es6
@@ -21,18 +21,31 @@
       this.renderCustomerAge();
     }
 
+    padNumber(val) {
+      return (`00${val}`).substr(-2, 2);
+    }
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/customer-age.es6
```diff
@@ -21,18 +21,31 @@
       this.renderCustomerAge();
     }
 
+    padNumber(val) {
+      return (`00${val}`).substr(-2, 2);
```
AndrewVos: Hah this is sneaky, I like it!

benlovell:
Tres bien.

PR #297, opened 2016-11-23
Run js:lint using yarn where possible

AndrewVos:
I'm using [yarn](yarnpkg.com) locally, so allow for running `yarn run` if it exists, but falling back to `npm run`

### DIFF
--- a/lib/tasks/js.rake
+++ b/lib/tasks/js.rake
@@ -1,6 +1,12 @@
 namespace :js do
   desc 'lint all js'
   task :lint do
-    sh 'npm run js'
+    bin = command?('yarn') ? 'yarn' : 'npm'
+    sh "#{bin} run js"
+  end
+
+  def command?(name)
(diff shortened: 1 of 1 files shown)


benbarnett:
Might we get inconsistent results with this approach?

benlovell:
Can you comment on why you're doing this?

AndrewVos:
@benbarnett We shouldn't, because for anyone running npm we use `npm run`, and for anyone running yarn we use `yarn run`. `npm` is default so you all should be ok.

@benlovell Updating description.

benbarnett:
But for those that do have yarn, can we be sure eslint will run in the same way? What if the versions diverge etc? I guess it's nothing too major and unlikely, though.

AndrewVos:
@benbarnett gotcha. That's true. I could check in the yarn lock file but then yarn would be consistently versioned and npm wouldn't :/

PR #305, opened 2016-11-24
Make the utilisation report show multiple days

AndrewVos:
- Date range is now required in utilisation reports

### DIFF
--- a/app/forms/utilisation_report.rb
+++ b/app/forms/utilisation_report.rb
@@ -11,31 +11,35 @@ class UtilisationReport
   end
 
   def generate
-    bookable, blocked = bookable_and_blocked
     CSV.generate do |csv|
       csv << [:booked_appointments, :bookable_slots, :blocked_slots]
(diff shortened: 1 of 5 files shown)


app/forms/utilisation_report.rb
```diff
@@ -11,31 +11,39 @@ def initialize(params = {})
...
+    [
+      day,
+      Appointment.where(start_at: day_range).count,
+      bookable,
+      blocked
+    ]
+  end
+
+  def bookable_and_blocked(day_range)
+    slots_within_range = slots_within_range(day_range)
```
benlovell: Maybe call this `slots` or something as it's a bit weird in that it shadows the method at first glance.

AndrewVos: Yeah true, and on second though I'm gonna remove `slots_within_range` and inline it. It's not really doing enough to warrant a separate method.

PR #306, opened 2016-11-24
Adding 'quick search' functionality

Guntrisoft:
screen shot 2016-11-24 at 13 45 36

- Altered search to redirect to an appointment edit page
  if it's the only result returned
- Added quick search form in the site navigation which
  submits directly to the appointment search page
- Added reschedule and rebook buttons to appointment edit
  page, so search results page isn't the only place to access
  these actions

### DIFF
--- a/app/assets/stylesheets/components/_quick-search.scss
+++ b/app/assets/stylesheets/components/_quick-search.scss
@@ -1,0 +1,15 @@
+$quick-search-panel-width: 233px;
+
+.quick-search__dropdown {
+  box-sizing: border-box;
+  padding: 8px;
+  width: $quick-search-panel-width;
+}
+
+.quick-search__input {
+  font-weight: normal;
+}
+
+.quick-search__label {
+  margin-bottom: 0;
+}
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -30,7 +30,7 @@ class AppointmentsController < ApplicationController
     @appointment.assign_to_guider
     if @appointment.save
       Notifier.new(@appointment).call
-      redirect_to search_appointments_path, success: 'Appointment has been rescheduled'
+      redirect_to edit_appointment_path(@appointment), success: 'Appointment has been rescheduled'
     else
       render :reschedule
     end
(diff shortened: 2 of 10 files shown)


app/controllers/appointments_controller.rb
```diff
@@ -66,11 +66,18 @@ def search
       search_params[:q],
       search_params[:date_range]
     )
+
+    check_for_exact_match
```
benlovell: I'd shuffle this around a bit. Firstly `redirect_on_exact_match` makes more sense as it's a command not a query. Secondly I'd pass `@search.results` as an argument to this method and move the cardinality check up to here, that was we can return early, otherwise the code below still runs:

```ruby
return redirect_on_exact_match(@search.results) if @search.results.count.one?

def redirect_on_exact_match(results)
  redirect_to #...
end
```

app/assets/stylesheets/components/_quick-search.scss
```diff
@@ -0,0 +1,9 @@
+.dropdown-menu--quick-search {
+  box-sizing: border-box;
+  padding: 8px;
+  width: 228px;
+}
+
+.quick-search-input {
+  width: 152px;
```
benbarnett: Same here

app/views/appointments/edit.html.erb
```diff
@@ -3,8 +3,22 @@
...
-<%= rebooked_from_heading(@appointment) %>
+<div class="row">
+  <div class="col-md-10">
+    <h1>Edit appointment for <%= @appointment.name %> <span class="text-muted"><span class="sr-only">Booking reference </span>#<%= @appointment.id %></span></h1>
+    <%= rebooked_from_heading(@appointment) %>
+  </div>
+  <div class="col-md-2 action-buttons">
+    <%= link_to appointment_reschedule_path(@appointment), title: 'Reschedule appointment', class: 'btn btn-info' do %>
+      <span class="glyphicon glyphicon-calendar" aria-hidden="true"></span>
+      <span class="sr-only">Reschedule appointment</span>
```
benbarnett: I'd be half tempted to show the full label of these buttons on this screen. Might be worth a try if it doesn't break the layout too badly?

app/assets/stylesheets/components/_quick-search.scss
```diff
@@ -0,0 +1,9 @@
+.dropdown-menu--quick-search {
+  box-sizing: border-box;
+  padding: 8px;
+  width: 228px;
```
Guntrisoft: As agreed, just the one pixel width value now

app/controllers/appointments_controller.rb
Guntrisoft: done

PR #307, opened 2016-11-24
Improve text wrapping on activity feed items

Guntrisoft:
**Before**
screen shot 2016-11-24 at 14 16 26

**After**

screen shot 2016-11-24 at 14 15 50

- Activity items with a large amount of text wrapped awkardly
- This fixes this issue by laying out an individual activity feed
  item into seperate columns


benbarnett:
Nice

PR #309, opened 2016-11-25
Move error element to page base class

benlovell:
This removes a bunch of duplication when requiring the authorisation
error message assertions. I made all the pages derive from the common
base class for consistency, whether they use this element or not.

I've had to explicitly require the base file in the spec helper since
the 'autoloading' glob loads the spec/support files in alphabetical
order, causing problems with load ordering on the pages that begin 'a*'.


AndrewVos:
Nice, I really like this.

PR #312, opened 2016-11-25
Improve layout of search form to bring results higher up the page

Guntrisoft:
**Before**
screen shot 2016-11-25 at 16 05 42

**After**
screen shot 2016-11-25 at 16 24 12

- making the form inline, so it fits on one row

### DIFF
--- a/app/views/appointments/search.html.erb
+++ b/app/views/appointments/search.html.erb
@@ -2,30 +2,36 @@
 
 <h1>Appointment search</h1>
 
-<div class="row form-group">
-  <div class="col-md-12">
-    <%= form_for @search, layout: :basic, method: 'GET', url: { action: :search } do |f| %>
-      <%= f.text_field :q, label: 'Search', class: 't-q', placeholder: 'Search by first name, last name, or booking reference.' %>
-      <%=
-        f.text_field(
-          :date_range,
-          class: 't-date-range',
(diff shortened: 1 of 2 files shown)


app/views/appointments/search.html.erb
```diff
@@ -2,30 +2,40 @@
...
-              autoUpdateInput: false,
-              locale: {
-                cancelLabel: 'Clear',
-                format: 'DD/MM/YYYY'
+<%= panel do %>
+  <div class="panel-body">
+    <%= form_for @search, method: 'GET', url: { action: :search }, html: { class: 'form-inline' } do |f| %>
+      <div class="form-group">
+        <%= f.label :q, 'Search term' %>
+        <%= f.text_field :q, use_label: false, class: 't-q appointment-search-term form-control', placeholder: 'Ref, Name or Guider' %>
```
benbarnett: Can we drop the caps & abbreviation on these: `Reference, name or guider`  should fit?

benbarnett:
Not sure how I feel about the border around the form. What do you reckon?

Guntrisoft:
@benbarnett removed the border

app/views/appointments/search.html.erb
Guntrisoft: done

PR #313, opened 2016-11-26
Fix date-sensitive spec

benlovell:
This spec was failing during the weekends due to its dependency on
working days.


Guntrisoft:
:+1:

PR #315, opened 2016-11-27
Use Ruby 2.3.3

benlovell:
This includes a fix (actually in 2.3.2 but this point release fixes a
regression) to open SSL to plug a significant leak.

Also includes many other fixes listed here:
http://svn.ruby-lang.org/repos/ruby/tags/v2_3_2/ChangeLog

Heroku / Travis are already compatible.


AndrewVos:
Sorry I don't have slack on my phone but in future if you fix a security issue feel free to text me!

benlovell:
It's not an urgent security issue as such. This doesn't actually fix a CVE so it was nothing to worry about. Thanks for merging though. 👌

PR #316, opened 2016-11-27
Reduce build queue times

benlovell:
In theory, using the lesser-used trusty distribution routing should
reduce time spent in the build queue. This won't have any further,
noticeable effect on our test runs.


AndrewVos:
Good thinking, hope it works

PR #319, opened 2016-11-28
Stop users from losing changes to sortable guiders

Guntrisoft:
screen shot 2016-11-28 at 12 00 03

- Add an action panel with the save button when ordering is changed
- If there has been a change in the ordering it sets the unload
  event to warn users about navigating away without saving changes

### DIFF
--- a/app/assets/javascripts/modules/sortable_guiders.es6
+++ b/app/assets/javascripts/modules/sortable_guiders.es6
@@ -6,11 +6,54 @@
     start(el) {
       super.start(el);
 
+      this.actionPanel = $('.js-action-panel');
+      this.saveButton = $('.js-save');
+
       this.init();
+      this.bindEvents();
     }
 
     init() {
-      new Sortable(this.$el[0]);
+      this.sortable = new Sortable(this.$el[0], {
+        onUpdate: () => this.onListChange()
+      });
+
+      this.pageOrder = this.getOrder();
+    }
+
+    bindEvents() {
+      this.saveButton.on('click', () => {
+        this.clearUnloadEvent();
+      });
+
+      $.subscribe('sortable-guiders-update', () => this.onListChange());
+    }
+
+    getOrder() {
+      return JSON.stringify(this.sortable.toArray());
+    }
+
+    onListChange() {
+      const currentOrder = this.getOrder();
+
+      this.clearUnloadEvent();
+
+      if (this.pageOrder !== currentOrder) {
+        this.setUnloadEvent();
+        this.actionPanel.fadeIn();
+      } else {
+        this.actionPanel.fadeOut();
+      }
+    }
+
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/modules/sortable_guiders.es6
```diff
@@ -6,11 +6,55 @@
     start(el) {
       super.start(el);
 
+      this.saveWarningMessage = 'You have unsaved changes - Save, or undo the changes.';
```
benbarnett: Is it worth moving this up to base? I think we have it in a few places now.

Guntrisoft: Done

app/assets/javascripts/modules/sortable_guiders.es6
```diff
@@ -6,11 +6,54 @@
...
+      this.saveButton.on('click', () => {
+        this.clearUnloadEvent();
+      });
+    }
+
+    getOrder() {
+      return JSON.stringify(this.sortable.toArray());
+    }
+
+    hasListChanged() {
```
benbarnett: This is a bit of a weird method name for something that doesn't return anything.

How about just `onListUpdate` or something?

app/assets/javascripts/tap-base.es6
```diff
@@ -11,6 +11,8 @@ class TapBase {
       this.getElementConfig(),
       this.getCookieConfig()
     );
+
+    this.saveWarningMessage = 'You have unsaved changes - Save, or undo the changes.';
```
benbarnett: We'll probably want to remove this from the other modules as well, sorry

app/assets/stylesheets/components/_sortable-guiders.scss
```diff
@@ -0,0 +1,3 @@
+.sortable-guiders.changes {
+  margin-bottom: 38px;
```
benbarnett: Anything we can do to make this less magic?

app/assets/javascripts/tap-base.es6
benbarnett: Oh you did that already

benbarnett:
Those tests will need re-jigging slightly I think

app/assets/javascripts/tap-base.es6
Guntrisoft: ✅

app/assets/stylesheets/components/_sortable-guiders.scss
Guntrisoft: I've removed this - don't think it's worth doing really as you can still see the last item and drag it

benlovell:
I wonder if these specs may be a little timing-dependent on the panel showing now? Guess we'll find out 😄

PR #320, opened 2016-11-28
Stop multiple unload event alerts on allocations calendar

Guntrisoft:
screen shot 2016-11-28 at 14 19 34

- removed window.unload event as beforeunload is enough
  to work in Chrome/Firefox/IE11
- moved clear and set unload methods into TapBase

### DIFF
--- a/app/assets/javascripts/modules/calendars/allocations.es6
+++ b/app/assets/javascripts/modules/calendars/allocations.es6
@@ -309,20 +309,6 @@
       });
     }
 
-    clearUnloadEvent() {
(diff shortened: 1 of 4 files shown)


AndrewVos:
How does this fix the problem?

Guntrisoft:
@AndrewVos sorry - made commit and description clearer!

AndrewVos:
@Guntrisoft Oh I get it now, that's weird :)

AndrewVos:
(LGTM)

PR #321, opened 2016-11-28
Show BugMuncher on all environments and use environment variable

benbarnett:
Requires `BUGMUNCHER_KEY` to be an environment variable.

This is part of the change to show BugMuncher across all relevant environments, including production as we would like to get direct feedback from the users at TP and TPAS.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,10 +20,10 @@
   <% end %>
   </script>
 
-  <% if Rails.env.development? || Rails.env.staging? %>
+  <% if ENV['BUGMUNCHER_KEY'] %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -20,10 +20,10 @@
   <% end %>
   </script>
 
-  <% if Rails.env.development? || Rails.env.staging? %>
+  <% if defined? ENV['BUGMUNCHER_KEY'] %>
```
benlovell: You could drop the `defined?` here. If the key is present this evaluates as truthy.

benlovell:
🚢

PR #322, opened 2016-11-28
Restore missing flash message

benlovell:
We shouldn't be rendering the edit action after a successful update
Instead we should be following the POST / redirect / GET rule.
Consequently this means the expected flash message is displayed again.


Guntrisoft:
👍

PR #323, opened 2016-11-28
Show booking details on the edit appointment page

Guntrisoft:
screen shot 2016-11-28 at 15 51 45

- Displays appointment created date, appointment date/time
  and the current guider it is assigned to

### DIFF
--- a/app/views/appointments/edit.html.erb
+++ b/app/views/appointments/edit.html.erb
@@ -27,6 +27,35 @@
   )
 %>
 
+<hr>
+<div class="row">
+  <div class="col-md-12">
+    <h2>Booking details</h2>
+  </div>
+</div>
+<div class="row">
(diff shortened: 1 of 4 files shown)


app/views/appointments/edit.html.erb
```diff
@@ -27,6 +27,35 @@
...
+<hr>
+<div class="row">
+  <div class="col-md-12">
+    <h2>Booking details</h2>
+  </div>
+</div>
+<div class="row">
+  <div class="col-md-12">
+    <table class="table table-striped table-bordered">
+      <caption><span class="sr-only">Assignment details</span></caption>
```
benbarnett: Should probably make this text the same as the h2

Guntrisoft: Good spot - sorted 👍

PR #324, opened 2016-11-28
Highlight search term in Javascript

Guntrisoft:
screen shot 2016-11-28 at 17 41 06

- allow certain columns of the search results to
  have part of its contents replaced with a <mark>
  element to highlight a match

### DIFF
--- a/app/assets/javascripts/modules/search-highlight.es6
+++ b/app/assets/javascripts/modules/search-highlight.es6
@@ -1,0 +1,28 @@
+{
+  'use strict';
+
+  class SearchHighlight {
+    start(el) {
+      const term = this.getUrlParameter(encodeURIComponent('search[q]')),
+        regEx = new RegExp(`(${term})`, 'gi');
+
+      if (!term) {
+        return;
+      }
+
+      $(el).find('.js-allow-highlighting').each((index, td) => {
+        $(td).html($(td).html().replace(regEx, '<mark>$1</mark>'));
+      });
+    }
+
+    getUrlParameter(name) {
+      const replacedName = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]'),
+        regex = new RegExp(`[\\?&]${replacedName}=([^&#]*)`),
+        results = regex.exec(location.search);
+
+      return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+    }
+  }
+
+  window.GOVUKAdmin.Modules.SearchHighlight = SearchHighlight;
+}
--- a/app/assets/stylesheets/helper/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/helper/_bootstrap_overrides.scss
@@ -11,4 +11,5 @@ $mark-colour: #faf785;
 mark,
 .mark {
   background-color: $mark-colour;
+  padding: 0;
 }
--- a/app/models/appointment_search.rb
+++ b/app/models/appointment_search.rb
@@ -12,17 +12,6 @@ class AppointmentSearch
 
   private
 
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/modules/search-highlight.es6
```diff
@@ -0,0 +1,29 @@
+{
+  'use strict';
+
+  class SearchHighlight {
+    start(el) {
+      const term = this.getUrlParameter(encodeURIComponent('search[q]')),
```
benbarnett: Can we make the `term` slightly more configurable?

benbarnett: Ignore this

app/assets/javascripts/modules/search-highlight.es6
```diff
@@ -0,0 +1,29 @@
...
+        return;
+      }
+
+      $(el).find('.js-allow-highlighting').each((index, td) => {
+        $(td).html($(td).html().replace(regEx, '<mark>$1</mark>'));
+      });
+    }
+
+    getUrlParameter(name) {
+      name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
```
benbarnett: Surprised this passes eslint actually. It's generally a bad idea to manipulate the arguments directly, can we assign this to a new variable?

Tempted to enable this too: http://eslint.org/docs/rules/no-param-reassign

app/assets/javascripts/modules/search-highlight.es6
```diff
@@ -0,0 +1,29 @@
...
+
+      const regex = new RegExp(`[\\?&]${name}=([^&#]*)`),
+      results = regex.exec(location.search);
```
benbarnett: I know we discussed this at the beginning, but I really don't like this declaration style. Indenting the variables below the first `const` would make things so much more readable. 

Perhaps we can address this in another PR. I think we should also revisit the file naming convention to match the class they return, and make consistent how we deal with `$variables`.

AndrewVos: Yes please! It hurts my brain having them on the same column!

app/assets/javascripts/modules/search-highlight.es6
AndrewVos: Might avoid a potential injection vuln. in the future:

```js
var $mark = $('<mark></mark>');
$mark.text($1);
```

app/assets/javascripts/modules/search-highlight.es6
benbarnett: Just for completion I've added a trello card for these sorts of things here: https://trello.com/c/mmTSxhYa/244-javascript-style-consistencies

app/assets/javascripts/modules/search-highlight.es6
AndrewVos: Just looked at this and there's no vulnerability because you use `$1`, whooops..

app/assets/stylesheets/helper/_bootstrap_overrides.scss
```diff
@@ -11,4 +11,5 @@ $mark-colour: #faf785;
 mark,
 .mark {
```
AndrewVos: We probably don't, but thought it would be better if we override bootstrap properly in case someone adds a `.mark` and it's a different colour.

AndrewVos:
@Guntrisoft I've removed all the redundant ruby stuff. LGTM :)

app/assets/javascripts/modules/search-highlight.es6
Guntrisoft: done!

PR #325, opened 2016-11-29
Display a message if there is no activity on the 'My activity' page

Guntrisoft:
screen shot 2016-11-29 at 10 00 24

- Previously if there was no activity, the page was looking rather
  empty

### DIFF
--- a/spec/features/guider_views_their_activities_spec.rb
+++ b/spec/features/guider_views_their_activities_spec.rb
@@ -9,6 +9,14 @@ RSpec.feature 'Guider views their activities' do
     create(:guider)
   end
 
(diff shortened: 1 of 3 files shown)


AndrewVos:
Beautiful!

spec/features/guider_views_their_activities_spec.rb
```diff
@@ -9,6 +9,14 @@
     create(:guider)
   end
 
+  scenario 'No activities' do
+    given_a_browser_session_for guider do
+      and_they_have_no_appointments
```
benlovell: We can drop this step.

PR #328, opened 2016-11-29
Users can add and edit single or multiple day holidays

AndrewVos:
- Add a list view for holidays
- Display holidays as allDay *only* in the holidays calendar
- Don't concat guider names in holiday titles

image

image

image

image


app/assets/javascripts/modules/holiday-types.es6
```diff
@@ -0,0 +1,39 @@
+/* global TapBase */
+{
+  'use strict';
+
+  class HolidayTypes extends TapBase {
```
benbarnett: Got a few thoughts on this bit but we can revisit when we take a look at the UI stuff for this page? cc @Guntrisoft

Guntrisoft: Agreed

Guntrisoft:
👍 apart from a squash, looks good to me - epic work

app/controllers/holidays_controller.rb
```diff
@@ -18,44 +18,47 @@ def index
...
-    end
+  def batch_upsert
+    @holiday = BatchUpsertHolidays.new(
+      holiday_ids,
+      batch_create_params[:title],
+      batch_create_params[:all_day],
+      batch_create_params[:date_range],
+      user_ids
+    )
+    return redirect_to(holidays_path, success: 'Holiday has been updated') if @holiday.call
```
benlovell: Can you move the conditional above and add an else clause here? I think it would be clearer. Also it's minor but we tend not to parenthesise arguments to the more DSL-like methods (such as `redirect_to`) - could we perhaps drop the parens?

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def to_model
+    if @previous_holiday_ids.any?
+      @previous_holiday ||= Holiday.find(@previous_holiday_ids.first)
+    else
+      Holiday.new
+    end
+  end
+
+  def initialize(previous_holiday_ids = [], title = nil, all_day = true, date_range = nil, users = [])
```
benlovell: Can you move this under the `validates` section? Feels weird it being underneath another instance method.

benlovell: Also these arguments are begging to be named, not positional.

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+    @previous_holiday_ids = previous_holiday_ids
+    @title = title
+    @all_day = ActiveRecord::Type::Boolean.new.cast(all_day)
+    @date_range = date_range
+    @users = users
+  end
+
+  def call
+    return false unless valid?
+    Holiday.where(id: @previous_holiday_ids).destroy_all
```
benlovell: You can pass the conditions to `destroy_all` and get rid of the `where` here.

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+  def initialize(previous_holiday_ids = [], title = nil, all_day = true, date_range = nil, users = [])
+    @previous_holiday_ids = previous_holiday_ids
+    @title = title
+    @all_day = ActiveRecord::Type::Boolean.new.cast(all_day)
+    @date_range = date_range
+    @users = users
+  end
+
+  def call
+    return false unless valid?
```
benlovell: Put some breathing space under guard conditions.

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+    Holiday.where(id: @previous_holiday_ids).destroy_all
+    start_at, end_at = calculate_date_range
+    create_holidays(start_at, end_at)
```
benlovell: FWIW I feel the intermediary assignments of the start / end dates are not necessary. I'd favour nesting the call such as `create_holidays(calculate_date_range)`.

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    if all_day
+      parts.map { |d| strp_date_range_picker_date(d) }
+    else
+      parts.map { |d| strp_date_range_picker_time(d) }
+    end
+  end
+
+  def create_holidays(start_at, end_at)
+    User.where(id: users).each do |user|
```
benlovell: I wonder if we should be thinking about transacting these calls. Recovering from an interim failure might be tricky and not obvious for the user. This is more a question than a blocker.

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
...
+  attr_reader :users
+  attr_reader :all_day
+
+  validates :title, presence: true
+  validates :date_range, presence: true
+  validates :users, presence: true
+
+  def to_model
+    if @previous_holiday_ids.any?
+      @previous_holiday ||= Holiday.find(@previous_holiday_ids.first)
```
benlovell: Is it me or is it a bit surprising that you're memoizing based on a variable with a different name to the accessor?

app/forms/batch_upsert_holidays.rb
```diff
@@ -0,0 +1,63 @@
+class BatchUpsertHolidays
+  include ActiveModel::Model
+  extend ActiveModel::Naming
```
benlovell: Can we bunch the `includes` together on consecutive lines?

app/jobs/generate_bank_holidays_job.rb
```diff
@@ -5,17 +5,26 @@ def perform
...
 
   private
 
+  def create_holiday(title, start_at, end_at)
+    Holiday.find_or_create_by!(
+      title: title,
+      start_at: start_at,
+      end_at: end_at,
+      all_day: true,
+      bank_holiday: true
```
benlovell: All holidays are bank holidays?

config/routes.rb
```diff
@@ -28,8 +28,9 @@
     get :reschedule
   end
   resource :allocations, only: :show
-  resources :holidays, only: %i(index new create edit update) do
+  resources :holidays, only: %i(index new create edit) do
     post 'batch_create', on: :collection
```
benlovell: We could wrap all of these inside `collection do` and drop the `on: :collection` option.

db/migrate/20161125161211_add_all_day_to_holidays.rb
```diff
@@ -0,0 +1,10 @@
+class AddAllDayToHolidays < ActiveRecord::Migration[5.0]
+  def up
+    add_column :holidays, :all_day, :boolean, default: false, null: false
+    execute 'UPDATE holidays SET all_day = true WHERE bank_holiday = true'
```
benlovell: I know it's equivalent but I see no good reason for explicit SQL when `Holiday.where(all_day: true).update_all(bank_holiday: true)` infers for you. We should avoid SQL unless absolutely necessary.

spec/features/resource_manager_manages_holidays_spec.rb
```diff
@@ -92,64 +119,112 @@ def when_they_view_holidays
...
-      {
-        title: 'Merged Holiday - First Guider, Second Guider',
-        time: '13:00 - 14:00'
-      }
-    ]
+    holidays = @page.calendar.holidays
+
+    expect(holidays.count).to eq 2
+
+    expect(holidays.first[:title]).to eq 'First user holiday'
```
benlovell: You could push this into the page object, it gets called in unison frequently enough.

spec/features/resource_manager_manages_holidays_spec.rb
```diff
@@ -92,64 +119,112 @@ def when_they_view_holidays
...
 
     @page.create_holiday.click
     @page = Pages::NewHoliday.new
 
+    @page.create_a_multiple_day_holiday.set(true)
+    @page.title.set 'Holiday Title'
+    start_at = I18n.l(Time.zone.now.to_date, format: :date_range_picker)
+    end_at = I18n.l(Time.zone.now.to_date + 2.days, format: :date_range_picker)
+    @page.all_day_date_range.set "#{start_at} - #{end_at}"
+    page.execute_script '$(".daterangepicker").hide();'
```
benlovell: Can this not be executed in the context of the `@page`? We can push it into the page object as it's being used multiple times around these specs.

spec/features/resource_manager_manages_holidays_spec.rb
```diff
@@ -165,28 +240,48 @@ def and_there_are_guiders_with_holidays
     end
   end
 
-  def and_they_edit_the_holiday
+  def and_they_edit_the_multiple_day_holiday
```
benlovell: This step made me weep solemnly. The `start_at` `end_at` stuff is appearing elsewhere and a bunch of the other stuff too. We should try and push as much of this into the page as possible. The steps should be as imperative as we can make them.

spec/support/pages/holidays.rb
```diff
@@ -12,6 +12,8 @@ class Holidays < Base
       element :content, '.fc-content'
     end
 
+    section :calendar, Pages::CalendarSection, '.t-calendar'
```
benlovell: Makes no sense for a section to be namespaced within pages. Also suffixing the section with `Section` would not be necessary if it was namespaced within `Sections`.

app/controllers/holidays_controller.rb
AndrewVos: :heavy_check_mark:

app/controllers/holidays_controller.rb
```diff
@@ -18,44 +18,47 @@ def index
...
   end
 
   def create
     Holiday.create!(create_params)
   end
 
   def batch_create
-    @holiday = CreateHolidays.new(
+    @holiday = BatchUpsertHolidays.new(
+      nil,
```
AndrewVos: :heavy_check_mark:

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

AndrewVos:
```
DEPRECATION WARNING: Passing conditions to destroy_all is deprecated and will be removed in Rails 5.1. To achieve the same use where(conditions).destroy_all. (called from call at /home/vos/code/telephone_appointment_planner/app/forms/batch_upsert_holidays.rb:34)
```

app/forms/batch_upsert_holidays.rb
AndrewVos: :heavy_check_mark:

app/jobs/generate_bank_holidays_job.rb
benlovell: Rename the method `create_bank_holiday` perhaps?

db/migrate/20161125161211_add_all_day_to_holidays.rb
AndrewVos: Fair enough, I usually do this sort of thing so that we don't have to rely on class names in case they change in the future.

AndrewVos: :heavy_check_mark:

spec/features/resource_manager_manages_holidays_spec.rb
AndrewVos: :heavy_check_mark:

spec/features/resource_manager_manages_holidays_spec.rb
AndrewVos: :heavy_check_mark:

spec/support/pages/holidays.rb
AndrewVos: :heavy_check_mark:

spec/features/resource_manager_manages_holidays_spec.rb
AndrewVos: :heavy_check_mark:

PR #329, opened 2016-11-29
Relax client side valiation on date of birth fields

benbarnett:
The regular expression was be interpreted incorrectly on IE,
which was stopping TPAS or TP being able to book appointments
with certain birth dates.

So as long as we're checking for a numeric input then this should
suffice.


Guntrisoft:
slight typo in the commit description, otherwise 👍

benbarnett:
Whoops, fixed 👍

PR #334, opened 2016-11-29
Disallow creating holidays that overlap other holidays from allocations

AndrewVos:
I'm not sure is 100% what the acceptance criteria mentioned, but might be better?


Guntrisoft:
👍

PR #337, opened 2016-11-29
Prevent activity feed duplicates and stray websocket connections

benbarnett:
Stray socket connections were afoot.
These needed to get oot.
Time to pocket the unwanted sockets.
🚀

### DIFF
--- a/app/assets/javascripts/modules/activity-feed.es6
+++ b/app/assets/javascripts/modules/activity-feed.es6
@@ -17,10 +17,18 @@
     handlePushEvent(payload) {
       let $element = $(payload.body);
 
-      $element
-        .hide()
-        .prependTo(this.$el)
-        .fadeIn();
+      if (this.isUnique($element)) {
+        $element
+          .hide()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/activity-feed.es6
```diff
@@ -12,15 +12,33 @@
...
+          .fadeIn();
+      }
+    }
+
+    isUnique($element) {
+      const activityTimestamp = $element.data('timestamp'),
+        activityMessage = $element.find('.activity__message').text();
+
+      if ($(`.activity[data-timestamp='${activityTimestamp}']`).length &&
+          $(`.activity__message:contains('${activityMessage}')`).length) {
```
AndrewVos: You can use the activity ID here, it will be unique.

benbarnett: Much better. Have done.

benlovell:
LGTM. 🚢

PR #338, opened 2016-11-29
Fix broken appointment seeding

AndrewVos:
Appointments were failing to be assigned to guiders because we were
using appointments that were obscured by holidays.


Guntrisoft:
👍

PR #342, opened 2016-11-30
Add autoApply: true to any date pickers without time selection

Guntrisoft:
screen shot 2016-11-30 at 12 17 42

- When you select a date range it's nicer to close the date
  input dialog automatically so you can complete the form quicker


AndrewVos:
Nice

AndrewVos:
Real nice

PR #343, opened 2016-11-30
When deleting groups, destroy groups that aren't in use by anyone

AndrewVos:


app/forms/destroy_group_assignments.rb
```diff
@@ -8,6 +8,8 @@ def call
     GroupAssignment
       .where(user_id: user_ids, group_id: group_id)
       .delete_all
+    group = Group.find(group_id)
```
benlovell: Can you extract this into a private method like `destroy_orphaned_group(group_id)` perhaps? Also leave a bit of breathing space between the deletion of groups and the call.

benbarnett:
Yep this'll work 👍

app/forms/destroy_group_assignments.rb
AndrewVos: :checkered_flag:

benlovell:
Noice.

PR #346, opened 2016-11-30
Add a loading indicator to the calendars

Guntrisoft:
screen shot 2016-11-30 at 16 20 45

- on slow network connections it will be more obvious
  to the user that something is happening when they
  change the current view of the calendar


Guntrisoft:
@benbarnett fixed the issue some of the tests failing (was related to the schedule/new) page - used `hide` class to initially hide the loading view and then show it when events are being loaded.

benbarnett:
Cool, looks good. I get the feeling we're going to need to look at `_calendar.scss` soon, but probably post launch is safest 🙈

Guntrisoft:
@benbarnett yup, agreed!

PR #349, opened 2016-12-01
Altered way jquery finds DOM elements across different modules

Guntrisoft:
- GOVUK Admin standard is to use classes with .js- prefix
https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md
- altered any finding of DOM elements with $('[data-X]') to
  $('.js-X') to be consistent

### DIFF
--- a/app/assets/javascripts/modules/calendars/allocations.es6
+++ b/app/assets/javascripts/modules/calendars/allocations.es6
@@ -23,7 +23,7 @@
 
       this.eventChanges = [];
       this.isFullscreen = false;
-      this.$actionPanel = $('[data-action-panel]');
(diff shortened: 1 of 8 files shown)


AndrewVos:
Truly a noble cause.

benbarnett:
Please can you try and limit the force pushing after you've sent up for review? Makes it tricky to know what's changed

AndrewVos:
If you're sure this works then merge away :)

Guntrisoft:
I'm gonna do a manual run through of the things it affects and if all good i'll merge it!

benbarnett:
It needs a rebase to see the holiday stuff but seems like most things are working!

PR #350, opened 2016-12-01
Import Booking Bug appointments

benlovell:
This imports the existing booking bug appointments from the provided
CSV URI or file path.

Some things to note:

- The guider UIDs need to map correctly to staging / prod signon
- Appointments will be booked by the 'Pension Wise Import' agent
- If the guider isn't mapped they're marked `disabled`
- It's kinda slow but that's OK - it's an ad-hoc thing

Appointment creation is not validated since doing so would cause
problems with the two working day validations. As it stands we can be
pretty certain that the data from the CSV is good.

The specs are intentionally light since the nature of this import is
that it will need eyes on the imported data.


AndrewVos:
:100:

app/lib/appointment_row.rb
```diff
@@ -0,0 +1,54 @@
...
+    booking_reference: 40,
+    first_name: 41,
+    last_name: 42,
+    email: 43,
+    mobile: 54,
+    phone: 55,
+    memorable_word: 61
+  }.freeze
+
+  def initialize(row)
```
AndrewVos: One thing I might want to do here is just to raise when the row length is not what we expect. Just to be 100% sure.

app/lib/guider_importer.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  private
+
+  def mapped_attributes
+    { name: name }.tap do |attributes|
+      if mappings.key?(name)
+        attributes[:uid] = mappings[name]
+      else
+        attributes[:disabled] = true
```
benbarnett: Probably best we refer to these as `hidden`. Don't know if that needs to change here but just thought I'd point out.

app/lib/guider_importer.rb
```diff
@@ -0,0 +1,75 @@
...
+      'Michelle Cracknell'    => '1b2cf510-7355-0134-fe9d-5eea3e482a8b',
+      'Nick Mines'            => '1b5aac10-7355-0134-fe9d-5eea3e482a8b',
+      'Nicola Collinson'      => '1b89caf0-7355-0134-fe9d-5eea3e482a8b',
+      'Nigel Kern'            => '1bb3d320-7355-0134-fe9d-5eea3e482a8b',
+      'Pam Atherton'          => '1bdc6d60-7355-0134-fe9d-5eea3e482a8b',
+      'Philippa Aaronson'     => '1bfe3e00-7355-0134-fe9d-5eea3e482a8b',
+      'Rachel Ward'           => '1c17cf00-7355-0134-fe9d-5eea3e482a8b',
+      'Raj Kakkad '           => '1c3540e0-7355-0134-fe9d-5eea3e482a8b',
+      'Richard Church'        => '1c512f30-7355-0134-fe9d-5eea3e482a8b',
+      'Robert Dunmore '       => '1c6dcd10-7355-0134-fe9d-5eea3e482a8b',
```
benbarnett: I'm guessing these trailing spaces are intentional?

benlovell: Yeah, the tokens are aligned so it's easier to scan.

app/lib/guider_importer.rb
benlovell: This will require significant changes to signon and GDS-SSO. `disabled` is something internal to those. 😢

app/lib/appointment_importer.rb
```diff
@@ -0,0 +1,41 @@
+class AppointmentImporter
+  FAKE_DATE_OF_BIRTH = '1950-01-01'.freeze
```
benbarnett: One thing we _could_ do here, is infer a different age if `Are you aged between 50-54` is true. This could be handy as they're a different type of appointment.

app/lib/appointment_row.rb
benlovell: I'm doing that as a pre-requisite when handling the CSV. It does actually require a little clean-up before it's good to import. This needs documenting elsewhere.

app/lib/guider_importer.rb
benbarnett: Oh I thought we were controlling this in our app? Did I get confused earlier?

app/lib/guider_importer.rb
benbarnett: I mean in the name itself: `'Robert Dunmore '`

app/lib/appointment_importer.rb
benlovell: Yeah I did wonder about this but decided on this approach or sheer consistency of the imported data. The values for that column have a pretty low cardinality so the quality of stuff coming out from there is kinda low unfortunately.

PR #352, opened 2016-12-01
Adding a (C) after the text of an appointment in 'my appointments'

Guntrisoft:
screen shot 2016-12-01 at 15 53 31

- forcing visited link colour for the list view to be black and
  not purple, which is hard to see on the cancelled event bg


Guntrisoft:
@benlovell The alternative option would be to add the 'status' of the appointment to the event - although space is sometimes limited depending on what view of the calendar you are looking at.

PR #353, opened 2016-12-01
Break apart holiday fields into more logical fields

AndrewVos:
For multiple day holidays:

start_at (date)
end_at   (date)

For single day holidays:

date     (date)
start_at (time)
end_at   (time)

image

image


app/forms/batch_upsert_holidays.rb
```diff
@@ -38,26 +40,39 @@ def call
 
   private
 
-  def calculate_date_range
-    parts = date_range.split(' - ')
+  def calculate_range(options)
+    @start_at = options[:start_at] || Time.zone.now
+    @end_at = options[:end_at] || Time.zone.now
+
+    return if @start_at.is_a?(ActiveSupport::TimeWithZone) && @end_at.is_a?(ActiveSupport::TimeWithZone)
```
AndrewVos: I don't feel good about this part. Any thoughts on how to make it nicer?

dwhenry:
Have a few questions about how we are store all day events - i.e. are we store dates or datetimes? But it appears to all look pretty good.

Also thought I saw a chat message around changing the all day label to something more descriptive.

benbarnett:
Yeah can we call it "This holiday lasts for multiple days"?

AndrewVos:
@benbarnett Thanks, done.

AndrewVos:
@dwhenry We store a `start_at` and an `end_at` which are both datetime. All day events are basically the same as single day events.

PR #355, opened 2016-12-02
Correct confirmation email copy

AndrewVos:
image

image


app/views/appointment_mailer/confirmation.html.erb
```diff
@@ -1,36 +1,137 @@
...
+  <strong class="emphasize" style="font-size: 24px;">
+    <%= @appointment.start_at.to_time.to_s(:govuk_time) %>
+  </strong>
+</p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px 0;font-size: 16px;line-height: 1.315789474;">
+  Appointment lasts
+  <br>
+  <strong class="emphasize" style="font-size: 24px;">
+    <%= ((@appointment.end_at - @appointment.start_at) / 1.minutes).round %> minutes
```
benbarnett: Ah I've just noticed this isn't quite right.

Although we have 70 minute appointments in the database, the customer's appointment is only 45 minutes. We allow for more time so the guider has time to prepare either side.

It might be worth just hardcoding this in for now...

PR #358, opened 2016-12-02
Resource managers can book appointments any time after Time.zone.now

AndrewVos:


app/models/appointment.rb
```diff
@@ -27,7 +27,7 @@ class Appointment < ApplicationRecord
   validates :status, presence: true
   validates :guider, presence: true
 
-  validate :not_within_two_business_days
+  validate :not_within_two_business_days, unless: :agent_is_resource_manager?
```
AndrewVos: If the user is a guider (or any other type of user) then we need to validate that the appointment starts two business days from now.

AndrewVos: (So in the case they're a guider then the validation will run)

benbarnett: Of course. I'll get my coat.

AndrewVos: Haha, I think maybe it's a bit complex though. Gonna try think how to make it sound better.

benlovell:
This is probably a good opportunity to drop the messy validation and conditionals from here and instead put them in form objects closer to the action?

AndrewVos:
@benlovell if I understand you correctly, you mean adding form objects for `CreateAppointment` and `UpdateAppointment`? I'm not sure this is a good idea because then we'll have to do the validation in both those objects?

benlovell:
@AndrewVos OK, I might have a better grasp on what needs doing on that front, so I'll pick it up separately.

PR #361, opened 2016-12-02
Change 'Rebook' icon to more of a repeat symbol

benbarnett:
Thoughts?

Before:
screen shot 2016-12-02 at 17 21 19
screen shot 2016-12-02 at 17 25 44

After:
screen shot 2016-12-02 at 17 21 05
screen shot 2016-12-02 at 17 25 39


AndrewVos:
LGTM, though if you want you could make the two `sr-only` texts the same "Rebook appointment"

benbarnett:
Leaving the label as-is as per discussion just now (the button label is already on the appointment page so you have the context there)

PR #362, opened 2016-12-02
Guiders can be deactivated

AndrewVos:
image

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -16,7 +16,7 @@ class UsersController < ApplicationController
   end
 
   def sort
-    @guiders = User.guiders.includes(:groups)
+    @guiders = User.guiders.active.includes(:groups)
   end
 
   def sort_update
@@ -29,5 +29,25 @@ class UsersController < ApplicationController
     end
 
     redirect_to :sort_users, success: 'Guider order has been updated'
+  end
+
+  def deactivate
+    toggle_activation
+    redirect_to users_path, success: 'Guider has been deactivated'
+  end
+
+  def activate
+    toggle_activation
+    redirect_to users_path, success: 'Guider has been activated'
+  end
+
+  private
+
+  def toggle_activation
+    user = User.find(params[:user_id])
+
+    user.toggle!(:active)
+
+    GenerateBookableSlotsForUserJob.perform_later user
   end
 end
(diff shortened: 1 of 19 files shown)


config/routes.rb
```diff
@@ -14,6 +14,8 @@
     resources :schedules
     get 'sort', on: :collection
     post 'sort', on: :collection, action: :sort_update
+    post 'deactivate'
```
benlovell: To respect HTTP semantics these should be `patch`.

app/views/users/index.html.erb
```diff
@@ -41,7 +42,33 @@
                 <span class="label label-info <%= dom_id(group) %> t-group"><%= group.name %></span>
               <% end %>
             </td>
-            <td nowrap="true"><%= link_to "/groups?user_ids=#{guider.id}", title: "Manage groups for #{guider.name}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-tag" aria-hidden="true"></span><span class="sr-only">Manage groups for #{guider.name}</span><% end %> <%= link_to edit_user_path(guider), title: "Manage schedules for #{guider.name}", class: 'btn btn-info' do %><span class="glyphicon glyphicon-calendar" aria-hidden="true"></span><span class="sr-only">Manage schedules for #{guider.name}</span><% end %></td>
+            <td>
+              <% if guider.active? %>
+                <span class="glyphicon glyphicon-ok t-active-icon" aria-hidden="true"></span>
+                <span class="sr-only">User is active</span>
```
benlovell: Is this a 'User' or 'Guider'?

benlovell: Is this a 'User' or 'Guider'?

app/controllers/users_controller.rb
```diff
@@ -30,4 +30,18 @@ def sort_update
 
     redirect_to :sort_users, success: 'Guider order has been updated'
   end
+
+  def deactivate
+    user = User.find(params[:user_id])
```
benlovell: Maybe extracting lines 35-37 into something like this might remove some of that duplication and tidy this up a teeny bit?

```ruby
def toggle_activation
  user = User.find(params[:user_id])
  user.toggle!(:active)

  GenerateBookableSlotsForUserJob.perform_later(user)
end
```

PR #363, opened 2016-12-02
Don't redefine `#no_show?`

benlovell:
This is defined for free with Rails enums. You also get the
bang-equivalent for transitioning to that particular state. FYI.


AndrewVos:
Hahaha wow. Nice one!

PR #367, opened 2016-12-05
Explain activate/deactivate guider actions in more detail

benbarnett:
Also shows the 'activate' (play) button in the success colour
to stand out a little more.

Before:
screen shot 2016-12-05 at 12 19 10

After:
screen shot 2016-12-05 at 12 16 40


Guntrisoft:
👍

PR #368, opened 2016-12-05
Send reminder emails to customers 24 hours before their appointment

AndrewVos:
image

image

image


app/jobs/appointment_reminders_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  def perform
+    Appointment.within_48_hours.each do |appointment|
+      send_reminder(appointment)
+    end
+  end
+
+  private
+
+  def send_reminder(appointment)
+    expected_reminder_count = appointment.audits.inject(0) do |count, audit|
```
benlovell: I wonder if this can be simplified just by asking the appointment if it has a reminder activity created within the same 48hr period? If it does then move along, otherwise we're good to send.

app/jobs/appointment_reminders_job.rb
```diff
@@ -0,0 +1,28 @@
...
+      keys = audit.audited_changes.keys
+      keys.include?('start_at') ? count + 1 : count
+    end
+    actual_reminder_count = ReminderActivity.where(appointment: appointment).count
+
+    return if actual_reminder_count == expected_reminder_count
+
+    AppointmentMailer.reminder(appointment).deliver_later
+    ReminderActivity.create!(
```
benlovell: We really ought to do these the other way around since the mailer is called asynchronously (and has chance to recover from failures) but the persistence code doesn't.

app/jobs/appointment_reminders_job.rb
```diff
@@ -0,0 +1,28 @@
+class AppointmentRemindersJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Appointment.within_48_hours.each do |appointment|
+      send_reminder(appointment)
```
benlovell: I wonder if it makes sense for this job to fan out individual jobs per appointment. What happens if one (or more) fails here?

app/models/appointment.rb
```diff
@@ -16,6 +16,8 @@ class Appointment < ApplicationRecord
 
   belongs_to :rebooked_from, class_name: Appointment
 
+  scope :within_48_hours, -> { where('start_at < ?', 2.days.from_now) }
```
benlovell: Perhaps naming this `within_reminder_window` or something that hints at *why* it's 48 hours could be helpful?

app/jobs/appointment_reminders_job.rb
AndrewVos: Ah, good point.

benbarnett:
Argh sorry I updated the card title but not the AC – this should be 24 hours before, not 48

app/jobs/appointment_reminders_job.rb
AndrewVos: No this wouldn't work because what if someone reschedules an appointment in the same 24 hour period. In this case we *should* send another reminder email.

AndrewVos: Actually, no. Perhaps we shouldn't email them because they get an `updated` email?

AndrewVos: @benbarnett

benlovell: Exactly. I thought about the various issues and it seems to cover all the cases correctly as far as I could tell.

benbarnett: Yep, concur.

app/jobs/appointment_reminders_job.rb
```diff
@@ -0,0 +1,14 @@
+class AppointmentRemindersJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Appointment.pending.within_reminder_window.each do |appointment|
+      sent_in_current_period = ReminderActivity
+                               .where(appointment: appointment)
+                               .where(created_at: 12.hours.ago..12.hours.from_now)
```
benlovell: You could collapse the wheres and the any check into a single `exists?` with the conditions.

spec/jobs/appointment_reminders_job_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        travel_to(appointment.start_at - 5.hours) do
+          perform_job
+        end
+
+        expect(AppointmentMailer).to have_received(:reminder).with(appointment).twice
+        expect(mailer_double).to have_received(:deliver_later).twice
+
+        activity = appointment.activities.first
+        expect(activity).to be_a(ReminderActivity)
+        expect(activity.user).to eq current_user
```
benlovell: This is specified elsewhere too. Could remove these few assertions so any change would only break a single spec.

AndrewVos: Cool, I did this and also extracted a seperate test for the `ReminderActivity` creation.

app/jobs/appointment_reminders_job.rb
AndrewVos: :checkered_flag:

PR #369, opened 2016-12-05
Style appointments in the past in a lighter colour

Guntrisoft:
screen shot 2016-12-05 at 15 22 56

- Appointments in the past get a lighter background colour
- this includes cancelled appointments which have got a
  slightly different red colour which has the same
  saturation as a non-cancelled appointment
- the lighter colour helps the user get a visual sense
  of what appointments have now happened

### DIFF
--- a/app/assets/javascripts/modules/calendar.es6
+++ b/app/assets/javascripts/modules/calendar.es6
@@ -150,8 +150,12 @@ class Calendar extends TapBase {
 
   }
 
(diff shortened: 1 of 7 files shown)


benbarnett:
Sent to TPAS for review/opinions. Let's see.

Guntrisoft:
@benbarnett all good to merge now that TPAS have said it's good?

PR #370, opened 2016-12-05
Ensure that ajax loading graphic shows when calendars first load

Guntrisoft:
loading

- Altered where calendar loading view gets inserted, to ensure
  that loading graphic is shown when calendar is first rendered

### DIFF
--- a/app/assets/javascripts/modules/calendar.es6
+++ b/app/assets/javascripts/modules/calendar.es6
@@ -53,7 +53,6 @@ class Calendar extends TapBase {
 
     this.insertJumpToDate();
     this.addAccessibilityImprovements();
-    this.insertLoadingView();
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/calendar.es6
```diff
@@ -135,6 +134,10 @@ class Calendar extends TapBase {
   }
 
   loading(isLoading) {
+    if (!this.calendarLoadingView) {
```
AndrewVos: Could be nice to move this check into the `insertLoadingView` function:

```
if (this.calendarLoadingView) { return; }
```

Guntrisoft: Done

PR #374, opened 2016-12-05
Get appointments needing reminder from Appointment

AndrewVos:


spec/models/appointment_spec.rb
```diff
@@ -248,4 +248,58 @@
       end
     end
   end
+
+  describe '#needing_reminder' do
```
benlovell: The notation is off here for a class method.

PR #376, opened 2016-12-06
Don't allow holidays with backwards date ranges

AndrewVos:
Needs some work:

- [x] When failing validation in one type (for example `all_day: true`), and then switching to the other type, the validation errors are still shown
- [x] Validation should fail when no users are selected

image

image

### DIFF
--- a/app/controllers/holidays_controller.rb
+++ b/app/controllers/holidays_controller.rb
@@ -17,7 +17,10 @@ class HolidaysController < ApplicationController
   end
 
   def new
-    @holiday = BatchUpsertHolidays.new
+    @holiday = BatchUpsertHolidays.new(
+      start_at: Time.zone.now,
+      end_at: Time.zone.now
+    )
   end
 
   def create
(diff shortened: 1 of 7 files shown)


spec/forms/batch_upsert_holidays_spec.rb
```diff
@@ -102,6 +118,18 @@
         Time.zone.parse('2016-10-12 14:00:00.000000000 +0000')
       end
 
+      it 'validates end_at is after start_at' do
+        subject = described_class.new(
+          options.merge(
+            single_day_start_at:       '12/10/2016',
```
benlovell: This just looks hideous.

app/controllers/holidays_controller.rb
```diff
@@ -80,19 +83,20 @@ def create_params
...
-    params
-      .require(:holiday)
-      .permit(
-        :title,
-        :all_day,
-        :start_at,
-        :end_at
-      )
-      .merge(users: user_ids)
+    params.require(:holiday)
```
benlovell: What's the rationale for changing the indentation style here? Can we keep the original style instead? It's consistent.

app/controllers/holidays_controller.rb
```diff
@@ -80,19 +83,20 @@ def create_params
...
+            :single_day_end_at,
+            :multi_day_start_at,
+            :multi_day_end_at
+          )
+          .merge(users: user_ids)
   end
 
   def user_ids
-    params[:holiday][:users] || []
+    params[:holiday][:users].select(&:present?) || []
```
benlovell: This implies that that param is always an array right, empty or not. If so, `select` would return a `[]` anyway so the or clause is redundant.

AndrewVos: Yeah that's true. I added the select because I was getting `[""]`...

app/controllers/holidays_controller.rb
AndrewVos: Well I could have added the `rubocop:disable` or just moved that one line up. Thought the latter would look better.

benlovell:
:shipit:

benlovell:
@AndrewVos I think a rebase would be lovely here.

PR #379, opened 2016-12-07
Agents see an appointment preview before they can confirm

AndrewVos:
~~@benbarnett Not sure about that copy, so keeping in `[draft]` for now.~~

~~This is my third approach at doing this. The other two were quite ugly and used javascript :cry:~~

~~I'm sure there are some nice refactorings that could be done so I'm taking a look at that now. Feel free to review it now though.~~

image

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -50,10 +50,21 @@ class AppointmentsController < ApplicationController
     @appointment = Appointment.copy_or_new_by(params[:copy_from])
   end
 
+  def preview
+    @appointment = Appointment.new(create_params.merge(agent: current_user))
+    @appointment.assign_to_guider
+    if @appointment.valid?
+      render :preview
+    else
+      render :new
+    end
+  end
+
   def create
     @appointment = Appointment.new(create_params.merge(agent: current_user))
     @appointment.assign_to_guider
-    if @appointment.save
+
+    if creating? && @appointment.save
       AppointmentMailer.confirmation(@appointment).deliver_later
       redirect_to(search_appointments_path, success: 'Appointment has been created')
     else
@@ -130,5 +141,9 @@ class AppointmentsController < ApplicationController
 
   def update_reschedule_params
     params.require(:appointment).permit(:start_at, :end_at)
+  end
+
+  def creating?
+    params[:edit_appointment].nil?
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/appointments_controller.rb
```diff
@@ -50,10 +50,21 @@ def new
...
+      render :new
+    end
+  end
+
   def create
     @appointment = Appointment.new(create_params.merge(agent: current_user))
     @appointment.assign_to_guider
-    if @appointment.save
+
+    if !edit_appointment? && @appointment.save
```
benlovell: Can we remove the negation? If we switch the helper to something like this instead:

```ruby
def creating?
  params[:edit_appointment]
end
```

app/views/appointments/preview.html.erb
```diff
@@ -0,0 +1,80 @@
...
+    { title: 'Preview' },
+  )
+%>
+
+<h1>Preview appointment</h1>
+
+<%= rebooked_from_heading(@appointment) %>
+
+<div class="alert alert-warning">
+  This booking has not been created yet. Please confirm the details are correct before proceeding.
```
benlovell: We should say 'appointment' not 'booking' for consistency.

app/views/appointments/preview.html.erb
```diff
@@ -0,0 +1,80 @@
...
+  <%= f.hidden_field :last_name %>
+  <%= f.hidden_field :email %>
+  <%= f.hidden_field :phone %>
+  <%= f.hidden_field :mobile %>
+  <%= f.hidden_field :memorable_word %>
+  <%= f.hidden_field :notes %>
+  <%= f.hidden_field :opt_out_of_market_research %>
+  <%= f.hidden_field :status %>
+
+  <div class='t-preview'>
```
Guntrisoft: quotes

spec/features/agent_manages_appointments_spec.rb
```diff
@@ -190,6 +221,59 @@ def when_they_want_to_book_an_appointment
...
+  def when_they_go_back_to_edit_their_appointment
+    @page = Pages::PreviewAppointment.new
+    expect(@page).to be_displayed
+    @page.edit_appointment.click
+  end
+
+  def and_they_change_some_details
+    @page = Pages::NewAppointment.new
+
+    @page.first_name.set                 'Another'
```
benlovell: Not a huge fan of this alignment, to me at least it's actually more difficult to scan / read. Do you find it much easier? (not a blocker, just a question)

Guntrisoft:
Markup wise I think it's ok, although I might have gone with bold label's for each item

AndrewVos:
@Guntrisoft Changed to `<strong>`

PR #380, opened 2016-12-08
Fix test that relied on the date always being 7th December 2016

AndrewVos:
The date is sometimes other days too.


benlovell:
Lollerskates.

PR #381, opened 2016-12-08
Preload appointment relations

benlovell:
This speeds up the reporting for appointments significantly, when faced
with a large data set.

Also took an opportunity to make some small refactoring improvements to
the style / structure.


AndrewVos:
Nice. Might be worth adding `bullet` to catch some N+1s too.

benlovell:
@AndrewVos yeah, good shout. I'll add that separately.

PR #382, opened 2016-12-08
Reduce huge N+1 querying for associated user IDs

benlovell:
This was causing a huge N+1 query through groups / assignments / users.


AndrewVos:
:+1:

dwhenry:
good to go once green. :)

PR #384, opened 2016-12-08
Speed up appointments JSON responses

benlovell:
Let the browser cache the responses based on the given etags. These
responses are trivially cacheable and testing locally shows that paging
back and forth through days that trigger conditional 304s drops the
response times for this endpoint down to ~100ms from ~500ms.

The recent `#stale?` addition performs checks to determine the freshness
of the given model / model collection instance by counting the elements
and retrieving the `MAX("appointments"."updated_at")`, merging the
same `where` clauses for the given model / model collection.

Since we had no index on this column this query was very slow (at least
in postgresql terms) in the order of `~100ms`. Adding this index drops
this down to `~1ms` which is quite some gain.

### DIFF
--- a/app/assets/javascripts/modules/calendars/company.es6
+++ b/app/assets/javascripts/modules/calendars/company.es6
@@ -26,7 +26,8 @@ class CompanyCalendar extends Calendar {
       eventTextColor: '#fff',
       eventSources: [
         {
-          url: '/appointments'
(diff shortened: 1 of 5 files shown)


AndrewVos:
You may need to stop fullcalendar from cache busting. I think by default it sends a _ along.

PR #385, opened 2016-12-09
Cache bookable slots responses

benlovell:
Use conditional GETs for bookable slot responses. In order to generate the
cache digests, Rails needs timestamps on the models so I added those.

The changes surfaced a timing issue on guiders / resources so I also
addressed this issue.

Adding indexes on the start and end dates for both holidays and
bookable slots lends quite an improvement in querying response times across
the board since most queries require these clauses in unison.

### DIFF
--- a/app/controllers/bookable_slots_controller.rb
+++ b/app/controllers/bookable_slots_controller.rb
@@ -1,10 +1,8 @@
 class BookableSlotsController < ApplicationController
   def index
-    slots = BookableSlot
-            .without_holidays
-            .within_date_range(start_at, end_at)
-    slots = slots.for_guider(current_user) if scoped_to_me?
-    render json: slots
+    @bookable_slots = bookable_slots
+
+    render json: @bookable_slots if stale?(@bookable_slots)
   end
 
   def available
@@ -20,6 +18,15 @@ class BookableSlotsController < ApplicationController
   end
 
   private
+
+  def bookable_slots
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+
+    slots = slots.for_guider(current_user) if scoped_to_me?
+    slots
+  end
 
   def start_at
     params[:start].to_date.beginning_of_day
--- a/db/migrate/20161209113503_add_timestamps_to_bookable_slots.rb
+++ b/db/migrate/20161209113503_add_timestamps_to_bookable_slots.rb
@@ -1,0 +1,15 @@
+class AddTimestampsToBookableSlots < ActiveRecord::Migration[5.0]
+  class BookableSlot < ApplicationRecord; end
+
+  def change
+    add_timestamps :bookable_slots, null: true
+
+    BookableSlot.update_all(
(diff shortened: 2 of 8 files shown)


app/controllers/bookable_slots_controller.rb
```diff
@@ -21,6 +19,15 @@ def scoped_to_me?
 
   private
 
+  def bookable_slots
```
AndrewVos: I'm starting to think this should be a scope on `BookableSlot`, or at least renamed to something more obvious like `visible_bookable_slots`, which would explain why we aren't removing the slots obscured appointments.

As it is, these really aren't `bookable_slots` at all...

benlovell: Well at the least they are of *type* `BookableSlot`...

db/migrate/20161209113503_add_timestamps_to_bookable_slots.rb
```diff
@@ -0,0 +1,15 @@
+class AddTimestampsToBookableSlots < ActiveRecord::Migration[5.0]
+  class BookableSlot < ApplicationRecord; end
+
+  def change
+    add_timestamps :bookable_slots, null: true
+
+    BookableSlot.update_all(
+      created_at: Time.zone.now,
+      updated_at: Time.zone.now
+    )
```
AndrewVos: Should we store a reference to `Time.zone.now` up above, so that `created_at` and `updated_at` are exactly the same?

spec/support/sections/calendar.rb
```diff
@@ -13,6 +13,8 @@ def slots
     end
 
     def guiders
+      wait_until_resource_cells_visible
```
AndrewVos: Nice!

app/controllers/bookable_slots_controller.rb
```diff
@@ -21,6 +19,15 @@ def scoped_to_me?
...
+    slots = BookableSlot
+            .without_holidays
+            .within_date_range(start_at, end_at)
+
+    slots = slots.for_guider(current_user) if scoped_to_me?
```
AndrewVos: Maybe use a ternary here?

```
scoped_to_me? ? slots.for_guider(current_user) : scoped_to_me?
```

Hmm, maybe uglier though

benlovell: Yeah. Not sure that's any better.

db/migrate/20161209113503_add_timestamps_to_bookable_slots.rb
benlovell: Not sure that's of any importance really?

AndrewVos: I dunno it's just something I always think about when doing these timestamps migrations. Like if I wanted to find which ones were created by migration I would do something like `WHERE created_at == updated_at`.

Probably a dumb idea though. Ignore it...

AndrewVos:
:ship:

PR #386, opened 2016-12-10
Replace database cleaner with rewinder

benlovell:
Database rewinder is the new hotness. It's hands down faster than
database cleaner and doesn't require such complicated configuration.
Locally it seems to have shaved a good 10 seconds or more from my spec
runs.


AndrewVos:
😍

PR #390, opened 2016-12-12
Ensure guiders not present in signon are inactive

benlovell:
This makes sure that imported guiders, when not mapped from those accounts
present in Signon (as signified by their matching name / uid) are flagged
as both `disabled` and not `active`. `disabled` is terribly named but this
is unfortunately an attribute controlled by Signon.

This is to ensure the guiders do not appear under normal circumstances
throughout the planner but are still present to satisfy referential
integrity.


Guntrisoft:
👍

PR #391, opened 2016-12-13
Notify guiders from anywhere when appointments change

AndrewVos:
TPAs have requested that guiders can see appointment change notifications from any page, not just the My Appointments page. This change fixes that, and also ensures that the relevant calendars are updated in the process.

Also, fix [HTML injection vulnerabilities](https://en.wikipedia.org/wiki/Code_injection#HTML_script_injection) in the qtip rendering and notification modal.

### DIFF
--- a/app/assets/javascripts/modules/calendar-events.es6
+++ b/app/assets/javascripts/modules/calendar-events.es6
@@ -1,0 +1,50 @@
+/* global TapBase, Pusher */
+{
+  'use strict';
+
+  class CalendarEvents extends TapBase {
+    start(el) {
+      this.config = {};
+
+      super.start(el);
+      this.setupPusher();
+    }
+
+
+    setupPusher() {
+      const channel = Pusher.instance.subscribe('telephone_appointment_planner');
+
+      channel.bind(`${this.config.guiderId}`, this.handlePushEvent.bind(this));
+    }
+
+    handlePushEvent(payload) {
+      const $modal = $('#calendar-event-modal'),
+            $modalBody = $modal.find('.modal-body'),
+            $customer = $('<p class="t-customer"></p>'),
+            $start = $('<p class="t-start"></p>'),
+            $guider = $('<p class="t-guider"></p>');
+
+      $modalBody.empty();
+
+      $modalBody.append($('<p><b>Appointment updated</b></p>'));
+
+      $customer.text(`Customer: ${payload['customer_name']}`);
+      $modalBody.append($customer);
+
+      $start.text(`Date/Time: ${payload['start']}`);
+      $modalBody.append($start);
+
+      $guider.text(`Guider: ${payload['guider_name']}`);
+      $modalBody.append($guider);
+
+      $('.js-calendar-reload-on-events').each(function() {
+        $(this).fullCalendar('removeEvents');
+        $(this).fullCalendar('refetchEvents');
+      });
+      $modal.modal('show');
+    }
+
+  }
+
+  window.GOVUKAdmin.Modules.CalendarEvents = CalendarEvents;
+}
--- a/spec/features/guider_views_appointments_spec.rb
+++ b/spec/features/guider_views_appointments_spec.rb
@@ -42,6 +42,77 @@ RSpec.feature 'Guider views appointments' do
     end
   end
 
+  scenario 'Guider views their own appoinments, and is notified of appointment changes', js: true do
+    given_the_requisite_data
+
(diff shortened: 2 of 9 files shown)


app/assets/javascripts/modules/calendar-events.es6
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    handlePushEvent(payload) {
+      const $modal = $('#calendar-event-modal');
+
+      $modal.find('.modal-body').html(`
+        <p><b>Appointment updated</b></p>
+        <p class="t-customer">Customer: ${payload['customer_name']}</p>
+        <p class="t-start">Date/Time: ${payload['start']}</p>
+        <p class="t-guider">Guider: ${payload['guider_name']}</p>
```
AndrewVos: Hope we don't have any customers named `<script>alert("HEYYYY")</script>` ?

AndrewVos: Or single quotes or double quotes

AndrewVos: Ok the `<script>` thing is totally for real. Fixing.

spec/features/guider_views_appointments_spec.rb
```diff
@@ -42,6 +42,78 @@
     end
   end
 
+  context 'Appointment is updated by someone else', js: true do
+    scenario 'viewing their appointments' do
+      @guider = create(:guider)
```
benlovell: Why not extract these factory calls and assignments to something step-like (`given_the_requisite_background_data` or something) that is shared between this and the following scenario?

spec/features/guider_views_appointments_spec.rb
```diff
@@ -42,6 +42,78 @@
     end
   end
 
+  context 'Appointment is updated by someone else', js: true do
+    scenario 'viewing their appointments' do
```
benlovell: Can we make the combination of the context and scenario text follow on in proper English? Though ideally we'd just flatten the context into the two scenarios instead I guess. Don't worry about duplication it's incidental.

spec/features/guider_views_appointments_spec.rb
```diff
@@ -42,6 +42,78 @@
...
+    @page = Pages::MyAppointments.new
+
+    @page.wait_until_notification_visible
+
+    expect(@page.notification.customer.text).to include(@appointment.name)
+    expect(@page.notification.guider.text).to include(@other_guider.name)
+  end
+
+  def and_the_appointment_is_removed_from_their_calendar
+    expect(@page.calendar.events.count).to eq 0
```
benlovell: `expect(@page.calendar.events).to be_empty` reads nicer I think.

spec/support/sections/calendar.rb
```diff
@@ -4,6 +4,10 @@ class Calendar < SitePrism::Section
     elements :events, '.fc-event'
     elements :resource_cells, '.fc-resource-cell'
 
+    def appointments
+      background_events('appointment')
```
benlovell: Was thinking about how this is growing into something a little nasty. That JS stuff digs right down into the calendar internals rather than verifying on something closer to / the DOM elements. Just thinking out loud here and not necessarily a blocker but it would be nice to cut this down into something closer to the actual user experience, rather than the JS internals.

benlovell: Also these things have some super-duper timing issues too.

app/assets/javascripts/modules/calendar-events.es6
benlovell: Yikes.

spec/features/guider_views_appointments_spec.rb
AndrewVos: :heavy_check_mark: couldn't think of a good name so went with yours. Seriously wondered about using `given_a_guider_with_appointments_and_another_without_any`

AndrewVos: `given_some_guiders_one_of_which_has_an_appointment`

AndrewVos: `given_some_guiders_one_of_which_has_an_appointment_but_also_a_resource_manager`

spec/support/sections/calendar.rb
benlovell: I'll think about this separately and see what we can come up with.

AndrewVos:
@benlovell All changes made :eye:

PR #395, opened 2016-12-14
Tweaks to utilisation reports

AndrewVos:
Some changes have been requested to the way we count utilisation report data:

- Utilisation report slots don't count slots obscured by appointments
- Add :cancelled_appointments column to utilisation report
- Don't take into account cancelled appointments in the utilisation report
- Refactor the utilisation report spec because it wasn't testing the bookable slots properly, and was really ugly


app/forms/utilisation_report.rb
```diff
@@ -12,7 +12,7 @@ def initialize(params = {})
 
   def generate
     CSV.generate do |csv|
-      csv << [:date, :booked_appointments, :bookable_slots, :blocked_slots]
+      csv << [:date, :booked_appointments, :bookable_slots, :blocked_slots, :cancelled_appointments]
```
benlovell: Can we use the `%i(...)` notation here instead now?

app/models/appointment.rb
```diff
@@ -16,6 +16,9 @@ class Appointment < ApplicationRecord
 
   belongs_to :rebooked_from, class_name: Appointment
 
+  scope :cancelled, -> { where(status: [:cancelled_by_customer, :cancelled_by_pension_wise]) }
```
benlovell: Same with these scopes can we provide the `status` entries using the `%i(...)` literal notation?

PR #403, opened 2016-12-16
[DO NOT MERGE] Show late cancellations in utilisation report

AndrewVos:
Late cancellations are appointments that got cancelled with less than
two working days until the appointment starts.

### DIFF
--- a/app/forms/utilisation_report.rb
+++ b/app/forms/utilisation_report.rb
@@ -14,7 +14,7 @@ class UtilisationReport
 
   def generate
     CSV.generate do |csv|
-      csv << %i(date booked_appointments bookable_slots blocked_slots cancelled_appointments)
+      csv << %i(date booked_appointments bookable_slots blocked_slots cancelled_appointments late_cancellations)
       range.each do |day|
         csv << generate_for_day(day)
       end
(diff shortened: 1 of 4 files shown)


benlovell:
We can hold off on this for a while. Need to think about the efficiency of this report.

AndrewVos:
@benlovell I can remove the `[do-not-merge]` now yeah?

benlovell:
@AndrewVos Not yet. This report was already hosing the server, these changes make it even worse and we need to think of a smarter / more efficient approach to reporting generally.

I think maybe in the short term, a smarter apprach would be to start flagging appointments as 'late cancellations' in the background on a schedule, or in the foreground at the time they're cancelled. It seems like a concern that should be first-class to me anyway, regardless of the reporting needs.

At least this way we wouldn't be digging back into the audits, we're doing this more and more in several places and to be honest we misuse them a little.

AndrewVos:
Cool, I think I'll save CSV to `/tmp/blah` like this then: http://stackoverflow.com/questions/1517635/save-pl-pgsql-output-from-postgresql-to-a-csv-file#1517692

It should drop the memory usage massively.

AndrewVos:
:cry:

PR #406, opened 2016-12-18
Removes the N+1 querying when searching

benlovell:
The guiders association was being loaded after the fact, and causing an
N+1 query. This also occurred when querying with the `joins`
declaration, thus the need for `joins` and `includes`.


AndrewVos:
LGTM

PR #408, opened 2016-12-19
Retry jobs until activity is committed

benlovell:
This is a weird one. The activities should be committed synchronously
before the jobs are scheduled but somehow this isn't occurring.

This is a temporary fix to ensure they're at least being run. The jobs
are retried up to a maximum of 5 times, with a 1 second delay between
each. This should recover the `ActiveJob::DeserializationError`
failures (in most cases I'd expect it to be successful upon the first
retry).


AndrewVos:
LGTM

PR #411, opened 2016-12-19
Slots with holidays that have same start+end are excluded

AndrewVos:
Also, the test names now read better.

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -55,9 +55,10 @@ class BookableSlot < ApplicationRecord
   def self.without_holidays
     joins(<<-SQL
           LEFT JOIN holidays ON
+            -- The holiday is specifically for the user, or it is for everyone
             (holidays.user_id = #{quoted_table_name}.guider_id OR holidays.user_id IS NULL) AND
             (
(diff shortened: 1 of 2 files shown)


app/models/bookable_slot.rb
```diff
@@ -55,9 +55,10 @@ def self.without_appointments
   def self.without_holidays
     joins(<<-SQL
           LEFT JOIN holidays ON
+            -- The holiday is specifically for the user, or it is for everyone
             (holidays.user_id = #{quoted_table_name}.guider_id OR holidays.user_id IS NULL) AND
             (
-              (holidays.start_at < #{quoted_table_name}.start_at AND holidays.end_at > #{quoted_table_name}.end_at)
+              (holidays.start_at <= #{quoted_table_name}.start_at AND holidays.end_at >= #{quoted_table_name}.end_at)
               OR
               (holidays.start_at >= #{quoted_table_name}.start_at AND holidays.start_at < #{quoted_table_name}.end_at)
```
benlovell: Should this not be repeated here too?

AndrewVos: No because that would hide slots when a holiday *starts* at the *end* of a slot.

benlovell: Cool.

PR #414, opened 2016-12-20
We need to manually compile to favicon with newer versions of rails

AndrewVos:
https://github.com/alphagov/service-manual-publisher/commit/3a989ed465a656347f22950b63ffb46927b9a92e


benlovell:
As I said offline, we should fix this at source since it got fat-fingered in alphagov/govuk_admin_template#119

AndrewVos:
Totes, awaiting https://github.com/alphagov/govuk_admin_template/pull/140

PR #416, opened 2016-12-20
Alterations for the second import attempt

benlovell:
We have all the guiders already created and some of them have changed names
and such so we shouldn't try to create them when importing again.

Use the column indexed at 0 for the ID

This was incorrectly reading the `member ID` previously.


benlovell:
Self merged this one for tonight.

PR #419, opened 2016-12-20
Don't use gem bundled phantomjs binaries

benlovell:
Using homebrew's bottled binaries is so much faster on my local machine
when running the specs. I can only assume the `phantomjs-binaries` bundled
binary is built with weird or no optimisation flags.

This change alone yields at least a 20 second improvement on spec timings
for me and shouldn't impact anyone else. Just ensure you've the latest
phantomjs binary on your `PATH`.

Sometimes my tests were slow waiting for pusher's external asset to be served.
It has no bearing on the specs so we can blacklist it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,6 @@ source 'https://rubygems.org' do
     gem 'chronic'
     gem 'coveralls', require: false
     gem 'database_rewinder'
-    gem 'phantomjs-binaries'
     gem 'poltergeist'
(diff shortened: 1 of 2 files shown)


benlovell:
Hmm. Travis is caching our previous bundle and the `phantomjs` gem binary.

benlovell:
Cleared the caches and we're all good.

Guntrisoft:
👍

PR #421, opened 2016-12-21
Add missing cache control headers for assets

benlovell:
Adding these ensure public caches work properly for our assets.


Guntrisoft:
👍 💯

PR #422, opened 2016-12-21
Change activity copy to use proper sentences

AndrewVos:
Also, add icons for each activity.

image

### DIFF
--- a/app/jobs/appointment_reminder_job.rb
+++ b/app/jobs/appointment_reminder_job.rb
@@ -4,8 +4,7 @@ class AppointmentReminderJob < ApplicationJob
   def perform(appointment)
     ReminderActivity.create!(
       appointment: appointment,
-      owner: appointment.guider,
-      message: 'sent a reminder'
+      owner: appointment.guider
     )
     AppointmentMailer.reminder(appointment).deliver_later
   end
--- a/app/models/assignment_activity.rb
+++ b/app/models/assignment_activity.rb
@@ -1,16 +1,8 @@
 class AssignmentActivity < Activity
   ASSIGNED = 'assigned'.freeze
 
-  def message
-    if allocated?(super)
-      "#{owner.name} was allocated a new appointment"
-    else
-      "#{user_name} allocated this appointment to #{owner.name}"
-    end
-  end
-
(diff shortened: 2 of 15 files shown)


app/jobs/appointment_reminder_job.rb
```diff
@@ -5,7 +5,7 @@ def perform(appointment)
     ReminderActivity.create!(
       appointment: appointment,
       owner: appointment.guider,
-      message: 'sent a reminder'
+      message: 'sent an appointment reminder email to the customer'
```
benlovell: Maybe just worth keeping this as it was (maybe we should have just provided an empty string for `message` from the start) and putting this message directly in the template instead? My view on these is it's more important to use `message` when computing the message at runtime is potentially expensive - such as when we compute the audited fields.

AndrewVos: Absolutely. It scares me that some of this is hardcoded and some stored in the DB.

AndrewVos: I'm going to go through this code and remove the dependency on the `message` column, where appropriate.

app/jobs/appointment_reminder_job.rb
```diff
@@ -4,8 +4,7 @@ class AppointmentReminderJob < ApplicationJob
   def perform(appointment)
     ReminderActivity.create!(
       appointment: appointment,
-      owner: appointment.guider,
-      message: 'sent a reminder'
```
benlovell: For completion sake should we change the existing messages for this activity to `''`?

AndrewVos: This is old code you're looking at surely

AndrewVos: Oh disregard that I see what you mean. Message is defaulting to '' anyway so no reason for that.

AndrewVos: You mean in the DB right?

benlovell: Yep.

AndrewVos: I would rather not, out of fear

PR #426, opened 2016-12-21
Navigate to full host in phantom JS to work around bug in newer versions

AndrewVos:
https://github.com/ariya/phantomjs/issues/10632


AndrewVos:
Closed in favour of using prebuilt binaries from the phantomjs site. Turns out the package released under ubuntu is 2.1.1 but not really 2.1.1. It's a much older version that exhibits the above behaviour.

PR #428, opened 2016-12-21
Update listjs to latest version

Guntrisoft:
- 1.2.0 -> 1.3.0
- change history here: https://github.com/javve/list.js/blob/master/History.md


benbarnett:
Nice one

PR #429, opened 2016-12-21
Update daterangepicker to latest version

Guntrisoft:
- 2.1.21 -> 2.1.24
- see https://github.com/dangrossman/bootstrap-daterangepicker/releases for details


Guntrisoft:
@AndrewVos yeah it's a bit odd - however we are using a 1.X version with govuk_admin_template

PR #430, opened 2016-12-21
Iterate the appointments in batches

benlovell:
This may help a little with memory consumption when generating the
appointment report.


Guntrisoft:
👍

PR #431, opened 2016-12-22
Don't rely on GlobalID for these job arguments

benlovell:
Rails enlists an emplicit transaction around `save` or `update`
operations and callbacks fire within this enlisted transaction. We hook
into the audited gem's `after_audit` callback to enqueue pusher
notification jobs but this causes a race condition that in circa 20% of
the time results in the job relying on this data being dequeued before
the database transaction has committed said data.

Thanks to the unique way in which Active Job deserializes model-based
arguments to the jobs, an `ActiveJob::DeserializationError` is raised
that is not retried in the traditional sense.

Changing the job arguments to their ID references means we sidestep this
behaviour so when dependent pieces of data cannot be retrieved using
`Model.find` an `ActiveRecord::RecordNotFound` error is raised and our
retry behaviour steps in to enqueue another identical job in a few
seconds time - with almost certainty that the transaction committed in
the meantime. Each failure is also reported to bugsnag for
instrumentation.

This is a quick workaround until we can better schedule the pusher jobs
outside of `save`s enlisted transaction. We are currently seeing failed
notifications in production so this is being deployed to ensure all
notifications are eventually delivered.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ source 'https://rubygems.org' do
   gem 'audited'
   gem 'bh'
   gem 'bootstrap-kaminari-views'
+  gem 'bugsnag'
   gem 'font-awesome-rails'
   gem 'foreman'
   gem 'gds-sso'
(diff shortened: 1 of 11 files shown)


benlovell:
Smerging so we can get this into production.

PR #434, opened 2016-12-23
Adding a no-js message

Guntrisoft:
screen shot 2016-12-23 at 15 17 56

- hiding search and selection form elements from nojs browsers
- adding enable JS message for nojs browsers across various pages

### DIFF
--- a/app/views/shared/_nojs.html.erb
+++ b/app/views/shared/_nojs.html.erb
@@ -1,0 +1,5 @@
+<noscript>
+  <div class="alert alert-success" role="alert">
+    <p>Please ensure that Javascript is enabled on your browser.</p>
+  </div>
+</noscript>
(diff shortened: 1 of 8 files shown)


app/views/shared/_nojs.html.erb
```diff
@@ -0,0 +1,5 @@
+<noscript>
+<div class="alert alert-success" role="alert">
```
benlovell: Identation is a bit off here 😱 😉

app/views/shared/_nojs.html.erb
```diff
@@ -0,0 +1,5 @@
...
+  <p>Please ensure that javascript is enabled on your browser.</p>
```
benlovell: That should be 'JavaScript' right?

app/views/shared/_nojs.html.erb
Guntrisoft: fixed

app/views/shared/_nojs.html.erb
Guntrisoft: Correct! - fixed

PR #435, opened 2016-12-27
Remove Booking Bug CSV importer features

benlovell:
The import completed successfully so we don't need it around.


app/models/appointment.rb
```diff
@@ -1,5 +1,7 @@
 # rubocop:disable Metrics/ClassLength
 class Appointment < ApplicationRecord
+  FAKE_DATE_OF_BIRTH = '1900-01-01'.freeze
```
dwhenry: if we are moving this a more descriptive name might be nice. ```DATE_OF_BIRTH_FOR_IMPORTED_RECORDS```
might be a bit wordy however..

dwhenry:
LGTM.. just the me comment, feel free to ignore

PR #436, opened 2016-12-28
Ensure date of birth is at least 1900

benlovell:
The likelihood of this is pretty low but sticking with 1900-01-01 as a
cut-off for valid date of birth values will resolve the issues that
TPAS have raised.


Guntrisoft:
👍

PR #437, opened 2016-12-28
Correct pusher asset blacklisting during specs

benlovell:
This yields a good improvement as the specs aren't unnecessarily
waiting around for pusher's analytics assets. It does however highlight
even more timing issues in our specs that were being masked by the long
wait times for these assets.


Guntrisoft:
Retrospective 👍

PR #439, opened 2016-12-28
Open appointments in new tab to avoid losing position in calendar

Guntrisoft:
- added to company and allocations calendar

### DIFF
--- a/app/assets/javascripts/modules/calendars/company.es6
+++ b/app/assets/javascripts/modules/calendars/company.es6
@@ -67,6 +67,13 @@ class CompanyCalendar extends Calendar {
     $.get('/guiders', this.handleResources.bind(this, callback));
   }
 
+  eventClick(event) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/calendars/company.es6
```diff
@@ -67,6 +67,13 @@ class CompanyCalendar extends Calendar {
     $.get('/guiders', this.handleResources.bind(this, callback));
   }
 
+  eventClick(event) {
+    if (event.url) {
+      window.open(event.url);
+      return false;
```
benlovell: would `event.preventDefault;` be more suitable here?

Guntrisoft: Copied the example from fullcalendar's docs but preventDefault is fine in this case, so have changed it.

PR #441, opened 2016-12-29
Correct email presence check

benlovell:
We default a missing email to an empty string, so our reminder guard
should check for the same criterion.


Guntrisoft:
:+1:

PR #443, opened 2016-12-30
Upgrade all the things

benlovell:
Brings us up to date with our gem dependencies. The only weirdness was caused
by `poltergeist` and is related to this entry in their changelog:

```
Driver#evaluate_script/#execute_script no longer depend on the browser context
JSON object. Side effect of this is that objects returned with cyclic
references now have the cycles broken with nil rather than "(cyclic structure)" 
```

Check my commit messages for more details.


AndrewVos:
LG

PR #445, opened 2017-01-04
Include appointments from last day of date range in appointment reports

AndrewVos:


spec/features/resource_manager_downloads_appointment_reports_spec.rb
```diff
@@ -88,6 +96,19 @@ def then_they_get_reports_by_appointment_start_at
     expect_appointment_csv(@appointment_with_start_at)
   end
 
+  def then_the_appointment_is_included_in_the_result
```
benlovell: Can we change this to `and_` so the spec reads more naturally?

AndrewVos: Thanks, done

PR #446, opened 2017-01-06
Hide bookable slots that are partially overlapped by appointments

AndrewVos:
If an appointment start_at or end_at is BETWEEN the bookable slot
start_at and end_at, then hide the bookable slot.

@benbarnett @benlovell thoughts?

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -31,23 +31,27 @@ class BookableSlot < ApplicationRecord
       .without_holidays
   end
 
-  def self.without_appointments
+  def self.without_appointments # rubocop:disable Metrics/MethodLength
     joins(<<-SQL
             LEFT JOIN appointments ON
-              appointments.guider_id = #{quoted_table_name}.guider_id AND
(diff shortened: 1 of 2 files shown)


benlovell:
Perfecto 🐑

app/models/bookable_slot.rb
```diff
@@ -31,23 +31,27 @@ def self.bookable
...
                 #{Appointment.statuses['cancelled_by_pension_wise']}
               )
+              AND (
+                (appointments.start_at = #{quoted_table_name}.start_at AND appointments.end_at = #{quoted_table_name}.end_at)
+                OR
+                (appointments.start_at BETWEEN #{quoted_table_name}.start_at AND #{quoted_table_name}.end_at)
+                OR
+                (appointments.end_at BETWEEN #{quoted_table_name}.start_at AND #{quoted_table_name}.end_at)
+              )
             SQL
```
dwhenry: Believe this would achieve the same thing:
```
AND (
(appointments.start_at < #{quoted_table_name}.end_at 
AND appointments.end_at > #{quoted_table_name}.start_at)
)
```

PR #447, opened 2017-01-06
Add missing compound index on start / end at

benlovell:
We commonly query these fields together thus they benefit from a
compound index.


dwhenry:
Can we see in the explain plans, query run times that this improves performance? 😄

AndrewVos:
I've been wanting to add `spec/benchmarks` for a while :)

dwhenry:
Or get freaky and try something like: https://www.postgresql.org/docs/9.6/static/rangetypes.html

benlovell:
I've been playing with this locally and it seems adding a single index on `end_at` yields better results, which is absolutely unexpected. I'm going to close this for now while I figure this out. 👋

PR #448, opened 2017-01-06
Add a link to pre-populate the summary document

dwhenry:
* Allow guiders to open the summary document generator with data prepopulated
* add an endpoint so the summary document can trigger an activity entry once it has been created.

screenshot 2017-01-17 11 09 52

Need to set env variable:

```ruby
SUMMARY_DOCUMENT_GENERATOR
```

### DIFF
--- a/app/controllers/api/v1/summary_documents_controller.rb
+++ b/app/controllers/api/v1/summary_documents_controller.rb
@@ -1,0 +1,30 @@
+module Api
+  module V1
+    class SummaryDocumentsController < ActionController::Base
+      include GDS::SSO::ControllerMethods
+      before_action :require_signin_permission!
+
+      protect_from_forgery with: :null_session
+
+      def create
+        activity = SummaryDocumentActivity.new(summary_document_activity_params)
+        if activity.save
+          PusherActivityNotificationJob.perform_later(nil, activity.id)
+          head :created
+        else
+          render json: activity.errors, status: :unprocessable_entity
+        end
+      end
+
+      private
+
+      def summary_document_activity_params
+        {
+          appointment_id: params[:appointment_id],
+          owner: User.find_by(uid: params[:owner_uid]),
+          message: params[:delivery_method]
+        }
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/summary_documents_controller.rb
+++ b/app/controllers/api/v1/summary_documents_controller.rb
@@ -1,11 +1,6 @@
 module Api
   module V1
-    class SummaryDocumentsController < ActionController::Base
-      include GDS::SSO::ControllerMethods
-      before_action :require_signin_permission!
-
-      protect_from_forgery with: :null_session
-
+    class SummaryDocumentsController < Api::V1::ApplicationController
       def create
         activity = SummaryDocumentActivity.new(summary_document_activity_params)
         if activity.save
--- a/spec/requests/summary_documents_api_spec.rb
+++ b/spec/requests/summary_documents_api_spec.rb
@@ -1,0 +1,82 @@
+require 'rails_helper'
+
+RSpec.describe 'POST /api/v1/summary_documents' do
+  scenario 'create a valid summary document activity' do
+    given_the_user_is_a_pension_wise_api_user do
+      and_a_completed_appointment_exists
+      when_the_client_posts_a_valid_summary_document_request
+      then_the_service_responds_with_a_201
+      and_the_appointment_has_a_summary_document_activity_created
+    end
+  end
+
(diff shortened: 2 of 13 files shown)


dwhenry:
FYI, button location still needs to be considered, at the moment it is on the same row as the update button which is why the space below the status field.

config/initializers/summary_document_generator.rb
```diff
@@ -0,0 +1,5 @@
+SummaryDocumentGenerator.host = if Rails.env.development? || Rails.env.test?
```
benlovell: This isn't just the host. Perhaps we should rename it `uri`? Also, the check for dev or test environments is overkill in my opinion. We can just reduce it to `fetch` with the default you supplied across all environments. It will be apparent enough in production-like environments if it's misconfigured.

db/schema.rb
```diff
@@ -19,7 +19,7 @@
   create_table "activities", force: :cascade do |t|
     t.integer  "appointment_id",              null: false
     t.integer  "user_id"
-    t.string   "message",        default: ""
+    t.string   "message",        default: "", null: false
```
benlovell: How did this creep in? If this is down to someone borking the migrations / schema I think we should promote this to a separate PR.

app/helpers/summary_document_helper.rb
```diff
@@ -0,0 +1,12 @@
+module SummaryDocumentHelper
+  def link_to_summary_document(appointment)
+    return if appointment.no_show? || appointment.cancelled_by_customer? || appointment.cancelled_by_pension_wise?
```
benlovell: `#cancelled?` encapsulates the last two `cancelled_*` checks so you can shorten this a little. Also can you add a bit of whitespace under the guard condition so it's called out a little clearer?

app/helpers/summary_document_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+    if appointment.pending?
+      link_to 'Update status before creating summary document', '', class: 'btn btn-danger', disabled: true
```
benlovell: We can probably do with clarifying this text somehow. @benbarnett ?

app/lib/summary_document_generator.rb
```diff
@@ -0,0 +1,23 @@
+class SummaryDocumentGenerator
```
benlovell: I'm probably being a little picky but this isn't actually *generating* a summary document. Maybe `SummaryDocumentUrl` is more suited?

app/helpers/summary_document_helper.rb
benbarnett: Yeah we need to separate the instructions out away from the button label at the very least.

If we don't already, I think we should get a story written up for this so we can get the UI done properly.

app/controllers/api/v1/summary_documents_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        else
+          render json: activity.errors, status: :unprocessable_entity
+        end
+      end
+
+      private
+
+      def summary_document_activity_params
+        {
+          appointment: ::Appointment.find_by(id: params[:appointment_id]),
```
benlovell: You could just pass `appointment_id: params[:appointment_id]` here right?

spec/requests/summary_documents_api_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.describe 'POST /api/v1/summary_documents' do
+  include ActiveJob::TestHelper
+
+  scenario 'create a valid summary document activity' do
+    perform_enqueued_jobs do
```
benlovell: This (and the `include` above) isn't needed as the default adapter in test will run the jobs inline.

dwhenry:
@benlovell This should now (hopefully) be really for deployment on Tuesday.

spec/requests/summary_documents_api_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'rails_helper'
+
+RSpec.describe 'POST /api/v1/summary_documents' do
+  scenario 'create a valid summary document activity' do
+    given_the_user_is_a_pension_wise_api_user do
```
benlovell: One thing I just noticed - this is only working since that user has the signin permission, which technically means any user can access this API. We should limit this to the `PENSION_WISE_API_PERMISSION` to ensure only our API user can access this [per the example here](https://github.com/guidance-guarantee-programme/telephone_appointment_planner/blob/7f368c94e07b8d8ddb0ee65def2a5ce14f0cbbac/app/controllers/api/v1/appointments_controller.rb#L6)

PR #449, opened 2017-01-11
Add bookable slots API for customer booking

benlovell:
This adds a new JSON API endpoint at `/api/v1/bookable_slots` and
returns a payload like:

```
[
  "2017-01-01": [ "12:00", "13:00", "15:00" ],
  "2017-01-02": [ "12:10", "14:10" ]
]
```

It requires the `pension_wise_api` permission to be present for the
requesting user.

This first cut relies on grouping, de-duping and sorting in memory which
isn't ideal. The alternative turned into the beastiest SQL query I've
ever written and wasn't ideal for maintenance purposes, though a great
deal more performant.

Running this across the booking window (6 weeks) on a copy of the
production database, but in development mode yields the full response
in ~700ms.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,7 +28,9 @@ Style/FrozenStringLiteralComment:
 Metrics/MethodLength:
   Exclude:
     - 'spec/features/*'
+    - 'spec/requests/*'
 
 Metrics/AbcSize:
   Exclude:
     - 'spec/features/*'
+    - 'spec/requests/*'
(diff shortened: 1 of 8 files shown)


AndrewVos:
LGTM, though 700ms seems a bit scary!

benlovell:
@AndrewVos I know right. I'm certain I can work this down into something more respectable. 

It has also occurred to me that this endpoint doesn't need to be authed. It is public information and is essentially displayed verbatim on the calendar anyway. This will also cut out the need for a further hop between the frontend / backend for the main pension guidance project and the TAP backend. Removing that now.

PR #451, opened 2017-01-12
Don't rely on natural ordering

benlovell:
Ordering was busted since we weren't explicitly ordering before doing
the necessary transformations. We now order by the sliced up / grouped
start date for the bookable slot.


Guntrisoft:
👍

PR #452, opened 2017-01-12
Use the full timestamp in the slots response

benlovell:
To avoid unnecessary parsing tricks on the client we now need to return
the full ISO timestamp for the individual bookable slots.


benlovell:
cc @AndrewVos

PR #454, opened 2017-01-14
Introduce bookings API

benlovell:
This provides the basis for the bookings API at:

`POST /api/v1/appointments`

Callers will need the `pension_wise_api` permission from signon.

The form object is versioned using the same scheme as the API but since
the model and form classes are the same and Rails scoping is stupid,
I've had to dud alias the model class under a `Models` namespace so we
can qualify the scoping when calling the model `Appointment` from the
API `Appointment`.

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -31,5 +31,9 @@ FactoryGirl.define do
     factory :contact_centre_team_leader do
       permissions { Array(User::CONTACT_CENTRE_TEAM_LEADER_PERMISSION) }
     end
+
+    factory :pension_wise_api_user do
+      permissions { Array(User::PENSION_WISE_API_PERMISSION) }
+    end
   end
 end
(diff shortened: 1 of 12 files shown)


spec/factories/users.rb
```diff
@@ -31,5 +31,9 @@
     factory :contact_centre_team_leader do
       permissions { Array(User::CONTACT_CENTRE_TEAM_LEADER_PERMISSION) }
     end
+
+    factory :pension_wise_api do
```
AndrewVos: Maybe `_user` prefix because when creating one of these it wouldn't always be obvious:

`create(:pension_wise_api)`

AndrewVos:
Do you want to explicitly cover the case where a slot is unavailable?

AndrewVos:
Looks good to me other than two comments

benlovell:
Update the factory name. As we spoke about offline, no need to call out the slot validation explicitly since it's covered by the guider assignment.

PR #456, opened 2017-01-16
Ensure no slots are generated on weekends

AndrewVos:
- Delete slots that were created on Saturdays and Sundays.
- Stop adding weekend slots in JS
- Stop adding weekend slots in db:seed

http://momentjs.com/docs/#/get-set/day/

NOTE: Whoever merges this MUST run `rake jobs:bookable_slots:generate`.

### DIFF
--- a/app/assets/javascripts/modules/calendars/guider-slot-picker.es6
+++ b/app/assets/javascripts/modules/calendars/guider-slot-picker.es6
@@ -168,13 +168,15 @@
         eventsToAdd = [];
 
       for (let currentDate = moment(calendarStartDate); currentDate < calendarEndDate; currentDate.add(1, 'days')) {
-        for (let eventIndex in events) {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/modules/calendars/guider-slot-picker.es6
```diff
@@ -168,13 +168,15 @@
         eventsToAdd = [];
 
       for (let currentDate = moment(calendarStartDate); currentDate < calendarEndDate; currentDate.add(1, 'days')) {
-        for (let eventIndex in events) {
-          let currentEvent = events[eventIndex];
+        if (currentDate.day() != 0 && currentDate.day() != 6) {
```
benlovell: I feel we should call this logic out somewhere in one of the feature tests since we have no idea this works unless I boot the app up and step through manually.

PR #459, opened 2017-01-17
Accept opt-out preference in appointments API

benlovell:
Permit clients to send a customer's marketing preferences as part of the
API payload. This is then persisted in the appointment record.


AndrewVos:
LGTM

PR #461, opened 2017-01-17
Don't wrap appointment parameters in appointments api

AndrewVos:
We pass params to the API without wrapping them, so rails normally will
wrap the JSON in { "appointments" : JSON }. Disable this, so that
the data we receive looks like the data we send.


benlovell:
The fact this apparently changes constant resolution is weird and bugs me out. Thanks Rails. Shrug.

PR #463, opened 2017-01-17
[do-not-merge] When assigning a guider, search for appointments 70 mins long

AndrewVos:
Currently no appointments are 60 minutes long, so this code
would never work in production.

### DIFF
--- a/app/forms/api/v1/appointment.rb
+++ b/app/forms/api/v1/appointment.rb
@@ -32,7 +32,7 @@ module Api
       private
 
       def end_at
-        Time.zone.parse(start_at) + 1.hour
+        Time.zone.parse(start_at) + 70.minutes
       end
 
       def to_params # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


app/forms/api/v1/appointment.rb
```diff
@@ -32,7 +32,7 @@ def create
       private
 
       def end_at
-        Time.zone.parse(start_at) + 1.hour
+        Time.zone.parse(start_at) + 70.minutes
```
benbarnett: Should this `70.minutes` be extractorado somewhere elseio? 

I guess we could have one for how long we tell the users appointments last for (45), and how long they're actually allocated for in the system (70).

AndrewVos: Actually we don't even need the 70 minutes if we refactor how the query works. I opted for this route because it was faster, though.

I'm feeling like that was maybe the wrong choice...

AndrewVos: I'm gonna work on an alternative PR...

benlovell: What are you doing instead?

AndrewVos: @benbarnett Remove `end_at` entirely from the query...

AndrewVos: Sorry @benlovell not barnett

AndrewVos:
Closing in favour of https://github.com/guidance-guarantee-programme/telephone_appointment_planner/pull/464

PR #464, opened 2017-01-17
Only search for guider slots with start_at

AndrewVos:
Don't search for `end_at` too, because it doesn't really matter if we do
or don't. This also means we can search for a slot if we only know the
`start_at`, like in the case of the `appointments` `create` API.

### DIFF
--- a/app/forms/api/v1/appointment.rb
+++ b/app/forms/api/v1/appointment.rb
@@ -31,14 +31,9 @@ module Api
 
       private
 
-      def end_at
-        Time.zone.parse(start_at) + 1.hour
-      end
-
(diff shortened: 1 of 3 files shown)


app/forms/api/v1/appointment.rb
```diff
@@ -13,6 +13,7 @@ class Appointment
       attr_accessor :dc_pot_confirmed
       attr_accessor :opt_out_of_market_research
       attr_accessor :agent
+      attr_accessor :end_at
```
benlovell: Why do we need this now?

AndrewVos: We don't. It is removed.

PR #465, opened 2017-01-18
Use consistent dates in mailer templates

benlovell:
Prior to this change the date / time formatting was incorrect in the
confirmation email. As per other transactional emails - we split the
date and time to be displayed separately but this wasn't happening here.

Also - in the text part - we should indent the date header and value
since the whitespace here is meaningful and will be preserved by textual
clients.


Guntrisoft:
👍

PR #467, opened 2017-01-26
Reduce runtime for the slots API

benlovell:
Given the size and cartesian products of the various join filters required
to infer and limit the bookable slots, we were reducing down from multiple
millions of rows. Limiting the scope and range of the join filters speeds
this up from multiple seconds (across our working set) to just a few
milliseconds.

We can also rely on the database to order the slot times which also yields a
reasonable reduction in runtime.

I'll create further PRs reducing the ranges used on bookable slots elsewhere.

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -54,11 +54,9 @@ class BookableSlot < ApplicationRecord
                 #{Appointment.statuses['cancelled_by_pension_wise']}
               )
               AND (
-                (appointments.start_at = #{quoted_table_name}.start_at AND appointments.end_at = #{quoted_table_name}.end_at)
-                OR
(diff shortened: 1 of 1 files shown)


app/models/bookable_slot.rb
```diff
@@ -27,61 +27,65 @@ def self.find_available_slot(start_at)
...
             )
             SQL
          )
       .where('holidays.start_at IS NULL')
   end
 
+  def self.reduce_by_range(table, from, to)
+    return '' unless from && to
+
+    sanitize_sql(["AND (#{table}.start_at > ? AND #{table}.end_at < ?)", from, to])
```
dwhenry: I feel the `AND` not be in this method.

benlovell: This clause is included only when a range is provided further up the stack. It's not to be used like a scope or arel.

AndrewVos:
lgtm

PR #473, opened 2017-02-07
Show the event start time as you drag an appointment

Guntrisoft:
9aso81v9v0

Users were sometimes mistakenly changing the start time of
an appointment, because it was difficult to see what time
exactly they were dragging an appointment to.

This provides them the start time of where the item has
currently been dragged to, making it easier to spot
potential mistakes.

### DIFF
--- a/app/assets/javascripts/modules/calendars/allocations.es6
+++ b/app/assets/javascripts/modules/calendars/allocations.es6
@@ -32,7 +32,8 @@
 
       super.start(el);
 
-      this.$rowHighlighter = $(`<div class="calendar-row-highlighter"/>`).insertAfter(this.$el);
+      this.$rowHighlighter = $('<div class="calendar-row-highlighter"/>').insertAfter(this.$el);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/calendars/allocations.es6
```diff
@@ -193,11 +194,19 @@
           left: eventPosition.left,
           width: $timeRow.width()
         }).addClass('active');
+
+        this.$rowHighlighterTime.css({
+          top: eventPosition.top - 35,
```
benbarnett: Can we make this number a little less magic? Negative margin in the CSS maybe?

Guntrisoft: I've altered how it looks slightly to avoid having to use this number

benbarnett:
Is there any way you could do a screencast of this?

Guntrisoft:
@benbarnett done!

PR #476, opened 2017-02-09
Guard against rebooked pending appointments

benlovell:
We must ensure bookable slots are returned to the pool when an
appointment is rebooked.

screen shot 2017-02-09 at 12 14 43

screen shot 2017-02-09 at 12 15 06


Guntrisoft:
👍

PR #477, opened 2017-02-09
Allow users to add groups from an existing list or create a new one

Guntrisoft:
Previously users have had to remember the exact spelling of a group
in order to add users to existing one - sometimes creating
problems where a typo ends up creating a new tag

This uses the select2 tags feature to allow someone to select
an existing tag or tags, or create a new one


Guntrisoft:
@benlovell altered a `h4` to an `h2` to get the document outline correct - looks good to me otherwise 👍

PR #479, opened 2017-02-19
Log filtered API parameters with lograge

benlovell:
We were doing this for regular controllers but not for our API-based
controllers. This will help us debug issues that occur in staging and
production.


Guntrisoft:
:+1:

PR #483, opened 2017-03-04
Bump outdated gems

benlovell:
Tested OK in staging without the uglifier issues that plagued our last upgrade. 👍 

---

I had to pin `momentjs-rails` is the newest version of `momentjs` was
spewing lots of deprecation messages around our formatting of dates. We
can address this after.

This includes updates to all outdated gems by simply running `bundle
update`. The update includes the latest point release for Rails among
others.


Guntrisoft:
👍

PR #484, opened 2017-03-06
Notify slack on customer bookings

benlovell:
We should notify slack when a booking is received from the main Pension
Wise website, as part of the customer journey. This helps us keep track
of how many bookings are being made during the A/B test and gradual
rollout.


benlovell:
[As the dev team we want to be notified in slack when there has been a online telephone booking](https://trello.com/c/6DGB7tln/32-as-the-dev-team-we-want-to-be-notified-in-slack-when-there-has-been-a-online-telephone-booking)

Guntrisoft:
👍

PR #487, opened 2017-03-15
Change notification mails

jongilbraith:
* Can yank the letter opener commit if you're not keen on it.
* I didn't make the constant in `Appointment` an array of symbols as mentioned as it's being used to compare with `appointment.previous_changes.keys` which expects strings.
* The new email behaviour is covered by additional scenarios.
* The move of the cancelled/confirmed/missed mails into the job is covered by existing features.


Gemfile
```diff
@@ -69,6 +69,7 @@ source 'https://rubygems.org' do
   end
 
   group :development do
+    gem 'letter_opener'
```
benlovell: Can we remove this in favour of a mailer preview instead? This follows with the rest of our emails.

app/jobs/appointment_cancelled_notification_job.rb
```diff
@@ -0,0 +1,13 @@
+class AppointmentCancelledNotificationJob < ActiveJob::Base
+  queue_as :default
+
+  MESSAGE = 'cancelled'.freeze
```
benlovell: I think these message constants should be defined in the `CustomerUpdateActivity`. Any overriding reason they're defined here?

app/jobs/appointment_confirmed_notification_job.rb
```diff
@@ -0,0 +1,13 @@
+class AppointmentConfirmedNotificationJob < ActiveJob::Base
```
benlovell: Thinking about these job classes, maybe we should just collapse them and pass the `message` as an argument. We could then just `case` over the `message` and invoke the right there. The activity creation and push notification stuff would just work as they're common across all three. WDYT?

app/lib/notifier.rb
```diff
@@ -23,17 +23,18 @@ def guiders_to_notify(appointment)
...
     elsif appointment_missed?(appointment)
-      AppointmentMailer.missed(appointment).deliver_later
+      AppointmentMissedNotificationJob.perform_later(appointment)
     end
   end
 
-  def appointment_time_changed?(appointment)
-    appointment.previous_changes.slice('start_at', 'end_at').present?
+  def appointment_details_changed?(appointment)
+    appointment.previous_changes.any? &&
```
benlovell: WDYT about moving this to a predicate such as `Appointment#notify?` Then we could just call that in the `if` statement in `notify_customer`.

Gemfile
jongilbraith: 👍

app/jobs/appointment_cancelled_notification_job.rb
jongilbraith: Yep, makes sense.

app/lib/notifier.rb
jongilbraith: And do the same thing with `appointment_cancelled?` and `appointment_missed?` for consistency?

benlovell: Yeah good shout. Perhaps `notify_missed?` and `notify_cancelled?`?

app/jobs/appointment_confirmed_notification_job.rb
jongilbraith: Yeah, certainly would be nice to cut down on the boilerplate. I'll do a quick rejig and see how it looks.

app/lib/notifier.rb
jongilbraith: 👍

benlovell:
[As a user when I update certain customer details an updated confirmation email is sent to the customer](https://trello.com/c/JNDAAOYR/30-as-a-user-when-i-update-certain-customer-details-an-updated-confirmation-email-is-sent-to-the-customer)

PR #488, opened 2017-03-16
Suppress activity as well as mail when a past appointment’s details are updated

jongilbraith:


app/lib/notifier.rb
```diff
@@ -24,7 +24,9 @@ def guiders_to_notify
 
   def notify_customer
     if appointment_details_changed?
-      CustomerUpdateJob.perform_later(appointment, CustomerUpdateActivity::UPDATED_MESSAGE)
+      unless appointment.in_the_past?
```
benlovell: Given how this is used now it might be better expressed as `#future?` (or a name to that effect) and then you combine the two checks above like `if appointment_details_changed? && appointment.future?`

jongilbraith: I was in two minds on the naming of that, so a second opinion is very welcome. Shall change.

benlovell:
[As a user when I update certain customer details an updated confirmation email is sent to the customer](https://trello.com/c/JNDAAOYR/30-as-a-user-when-i-update-certain-customer-details-an-updated-confirmation-email-is-sent-to-the-customer)

PR #490, opened 2017-03-16
Make appointment search date range read only

jongilbraith:
Switch feature to js and use javascript to set the date
following warnings from Capybara about setting the value
of a read only field.


jongilbraith:
[Don't permit custom entered date ranges when searching](https://trello.com/c/HlzoY8xO/61-don-t-permit-custom-entered-date-ranges-when-searching)

PR #491, opened 2017-03-17
Change notification precedence

benlovell:
We were occasionally sending update emails when the status changes
along with other notifiable attributes. By simply changing the
precedence to check for status updates first we will always send the
correct notifications when the status changes.

### DIFF
--- a/app/lib/notifier.rb
+++ b/app/lib/notifier.rb
@@ -23,12 +23,12 @@ class Notifier
   end
 
   def notify_customer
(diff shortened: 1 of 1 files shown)


benlovell:
[Ensure the correct email notifications are sent when modifying / cancelling appointments](https://trello.com/c/6ysxiH5s/64-ensure-the-correct-email-notifications-are-sent-when-modifying-cancelling-appointments)

PR #494, opened 2017-03-27
Display timestamps respective of DST

benlovell:
This ensures timestamps are formatted with the correct offset for
'London'. Start and end timestamps for appointments, holidays, slots and
such are unaffected by this change as they're expressed in UTC without
an offset.


benlovell:
[Output dates / times with respect to DST](https://trello.com/c/iUmI2I7O/73-output-dates-times-with-respect-to-dst)

PR #495, opened 2017-03-27
Calculate grace period from close of play

benlovell:
We were previously calculating the two business days grace period from
the current date and time. We should have calculated this from the close
of play and appointments have inadvertently been created.


benlovell:
[Appointments cannot be created within the grace period of two whole working days](https://trello.com/c/gQsvkUd4/74-appointments-cannot-be-created-within-the-grace-period-of-two-whole-working-days)

PR #496, opened 2017-03-27
Revert "Calculate grace period from close of play"

benlovell:
Reverts guidance-guarantee-programme/telephone_appointment_planner#495

We may have jumped the gun a little on this one. Reverting until we can fully understand the issue with TPAS.


benlovell:
[Appointments cannot be created within the grace period of two whole working days](https://trello.com/c/gQsvkUd4/74-appointments-cannot-be-created-within-the-grace-period-of-two-whole-working-days)

PR #498, opened 2017-03-29
High priority activity

jongilbraith:
Notification badge
==============
screen shot 2017-03-28 at 16 57 23

Activity drop down
==============
*If the user is a guider, there is also a footer on the list linking to their activity feed. Agents don't have this, so there is no link for them.*
screen shot 2017-03-28 at 16 57 58

No high priority activity
=================
screen shot 2017-03-28 at 17 01 45

High priority item in activity feed
========================
screen shot 2017-03-28 at 16 58 37

Resolved high priority item
====================
screen shot 2017-03-28 at 17 00 40

Error resolving a high priority item
==========================
screen shot 2017-03-28 at 17 04 38

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -13,6 +13,16 @@ class ActivitiesController < ApplicationController
     end
   end
 
+  def resolve
+    @activity = Activity.find(params[:id])
+    @activity.resolve!(current_user)
+
+    respond_to do |format|
+      format.js { head :ok }
+      format.html { redirect_back(fallback_location: edit_appointment_path(@activity.appointment)) }
+    end
+  end
+
   private
 
   def message_params
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -7,7 +7,6 @@ class ActivitiesController < ApplicationController
 
   def create
     @activity = MessageActivity.create(message_params)
-    PusherActivityCreatedJob.perform_later(appointment.guider_id, @activity.id)
     respond_to do |format|
       format.js { render @activity }
     end
@@ -16,9 +15,6 @@ class ActivitiesController < ApplicationController
   def resolve
     @activity = Activity.find(params[:id])
     @activity.resolve!(current_user)
-
-    PusherHighPriorityCountChangedJob.perform_later(@activity.owner)
-    PusherHighPriorityCountChangedJob.perform_later(@activity.user) if @activity.user
 
     respond_to do |format|
       format.js { head :ok }
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -23,9 +23,8 @@ class ActivitiesController < ApplicationController
   end
 
   def high_priority
-    @activities = current_user
-                  .activities
-                  .high_priority
+    @activities = Activity
+                  .high_priority_for(current_user)
                   .unresolved
                   .order('id DESC')
                   .limit(5)
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -17,6 +17,9 @@ class ActivitiesController < ApplicationController
     @activity = Activity.find(params[:id])
     @activity.resolve!(current_user)
 
+    PusherHighPriorityCountChangedJob.perform_later(@activity.owner)
+    PusherHighPriorityCountChangedJob.perform_later(@activity.user) if @activity.user
+
     respond_to do |format|
       format.js { head :ok }
       format.html { redirect_back(fallback_location: edit_appointment_path(@activity.appointment)) }
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -22,6 +22,17 @@ class ActivitiesController < ApplicationController
     end
   end
 
+  def high_priority
+    @activities = current_user
+                  .activities
+                  .high_priority
+                  .unresolved
+                  .order('id DESC')
+                  .limit(5)
+
+    render layout: false
+  end
+
   private
 
   def message_params
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -7,7 +7,7 @@ class ActivitiesController < ApplicationController
 
   def create
     @activity = MessageActivity.create(message_params)
-    PusherActivityNotificationJob.perform_later(appointment.guider_id, @activity.id)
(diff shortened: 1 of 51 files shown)


jongilbraith:
[As someone who has booked an appointment via TAP I am notified when the confirmation email has bounced/dropped/failed](https://trello.com/c/briGD02R/28-as-someone-who-has-booked-an-appointment-via-tap-i-am-notified-when-the-confirmation-email-has-bounced-dropped-failed)

app/views/activities/_activity.html.erb
```diff
@@ -1,17 +1,33 @@
...
-  </li>
-<% end %>
+    <% if details %>
+      </a>
+    <% end %>
+    <%= render('activities/details', appointment: activity.appointment) if details %>
+    <span class="activity__date">
+      <%= activity.created_at.to_s(:govuk_date_short) %> (<%= time_ago_in_words(activity.created_at.in_time_zone('London'))%> ago)
+      <% if activity.high_priority? %>
+       · <span class="activity__priority<% if activity.resolved? %> activity__priority--resolved<% end %>">Priority</span>
```
benlovell: What's that dot at the start of this line?

spec/features/notifications_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  let(:agent) do
+    create(:agent)
+  end
+
+  let(:guider) do
+    create(:guider)
+  end
+
+  scenario 'A new high priority activity occurs' do
+    given_a_browser_session_for guider, agent do
```
benlovell: Can we parenthesise these user arguments everywhere we call this method? The way these steps are written makes it quite tricky to scan and pick them up.

app/controllers/activities_controller.rb
```diff
@@ -7,12 +7,31 @@ def index
 
   def create
     @activity = MessageActivity.create(message_params)
-    PusherActivityNotificationJob.perform_later(appointment.guider_id, @activity.id)
     respond_to do |format|
       format.js { render @activity }
     end
   end
 
+  def update_resolve
```
benlovell: Can we rename this action to just `resolve`?

app/controllers/activities_controller.rb
```diff
@@ -7,12 +7,31 @@ def index
...
+    @activity = Activity.find(params[:id])
+    @activity.resolve!(current_user) unless @activity.resolved?
```
benlovell: `resolve!` makes that guard check so we can just remove that postfix modifier and happily return earlier from `resolve!` I reckon.

app/jobs/pusher_high_priority_count_changed_job.rb
```diff
@@ -0,0 +1,20 @@
+class PusherHighPriorityCountChangedJob < ApplicationJob
+  queue_as :default
+
+  def perform(user_id)
```
benlovell: This happens outside of a transaction so we can safely pass the actual `User` instance here and let GlobalID do its thing. Thus removing the need for `User.find(...)` below.

app/models/activity.rb
```diff
@@ -1,10 +1,17 @@
...
+  HIGH_PRIORITY_ACTIVITY_CLASS_NAMES = %w(DropActivity).freeze
+
   belongs_to :appointment
   belongs_to :user
   belongs_to :owner, class_name: User
   belongs_to :prior_owner, class_name: User
+  belongs_to :resolver, class_name: User
   validates :owner, presence: true
 
+  scope :resolved, -> { where('resolved_at IS NOT NULL') }
```
benlovell: `where.not(resolved_at: nil)` reads nicer and is more portable.

app/models/activity.rb
```diff
@@ -1,10 +1,17 @@
...
+  scope :unresolved, -> { where('resolved_at IS NULL') }
```
benlovell: Likewise `where(resolved_at: nil)`

app/models/assignment_activity.rb
```diff
@@ -14,27 +15,31 @@ def self.from(audit, appointment)
...
       appointment: appointment,
       owner: appointment.guider,
       prior_owner_id: prior_owner.id
     )
+  end
 
-    PusherActivityNotificationJob.perform_later(appointment.guider_id, activity.id)
-    PusherActivityNotificationJob.perform_later(prior_owner.id, activity.id)
+  def pusher_notify_user_ids
+    case message
```
benlovell: For now this could just collapse to a ternary perhaps? `message == ASSIGNED ? owner_id : Array(owner_id, prior_owner_id)` - if you feel that reads better? I'm not _overly_ fussed though so feel free to ignore.

app/controllers/activities_controller.rb
```diff
@@ -7,12 +7,31 @@ def index
...
+
+    respond_to do |format|
+      format.js { head :ok }
+      format.html { redirect_to :back }
```
benlovell: Rails 5 has `redirect_back(fallback_location: ...)` that is better here.

spec/features/notifications_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'rails_helper'
```
benlovell: Can we have some feature coverage for resolving the activities too here?

app/controllers/activities_controller.rb
jongilbraith: Sure, I was following the pattern I saw elsewhere, so I'll update this and the other instance as well.

app/views/activities/_activity.html.erb
jongilbraith: Not sure, but a typo I imagine (@Guntrisoft?). I'll get rid of it...

Guntrisoft: It was designed to be a seperator

jongilbraith: 👍  shall reinstate

app/views/activities/_activity.html.erb
```diff
@@ -1,17 +1,36 @@
...
-  </a>
-<% else %>
-  </li>
-<% end %>
+    <% if details %>
+      </a>
+    <% end %>
+    <%= render('activities/details', appointment: activity.appointment) if details %>
+    <span class="activity__date">
+      <%= activity.created_at.to_s(:govuk_date_short) %> (<%= time_ago_in_words(activity.created_at.in_time_zone('London'))%> ago)
```
benlovell: This needs to be displayed with respect to the London time zone too. Sorry, missed this before.

PR #499, opened 2017-03-30
Ensure activities are only created when necessary

benlovell:
While the actual email was a noop thanks to the absence of a customer email
address on the associated appointment we were incorrectly creating
activities. This change ensures activities are only created (and skips a
heap of other machinery) when a customer email address is present.

### DIFF
--- a/app/lib/notifier.rb
+++ b/app/lib/notifier.rb
@@ -23,6 +23,8 @@ class Notifier
   end
 
   def notify_customer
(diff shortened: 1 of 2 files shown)


benlovell:
[Missed appointment activities should be created in response to the correct events](https://trello.com/c/erNeFLDh/88-missed-appointment-activities-should-be-created-in-response-to-the-correct-events)

PR #500, opened 2017-03-30
Default fields to avoid phantom audit messages

benlovell:
These fields are usually initialised to `nil` so when a form is bound
and submitted the values change to empty strings and are thus considered
'changed'. While technically true, the generated activity messages are
confusing since no noticeable change occurs to the user.

Initialising these fields to empty strings ensures that when the
comparison is made no changes are detected.


benlovell:
[When modifying an appointment some fields are incorrectly reported as modified](https://trello.com/c/abAyAAHm/65-when-modifying-an-appointment-some-fields-are-incorrectly-reported-as-modified)

PR #501, opened 2017-03-30
Allow for appointments being displayed in UTC

benlovell:
This is a temporary hack, please forgive me.

Appointments are stored / displayed in UTC but we're currently in BST.
An appointment scheduled for '12:20' actually takes place at '11:20'
according to the server, so when the appointment is updated after it
ends the `future?` check passes since '12:20' is greater than '11:20'.
If we substract an hour from the appointment's start time, we end up
with a true representation of time to compare with.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -127,7 +127,7 @@ class Appointment < ApplicationRecord
   end
 
   def future?
-    start_at > Time.zone.now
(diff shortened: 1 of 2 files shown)


benlovell:
[Output dates / times with respect to DST](https://trello.com/c/iUmI2I7O/73-output-dates-times-with-respect-to-dst)

PR #502, opened 2017-03-31
Allow management of appointment type

benlovell:
This gives agents and others the ability to specify the type of appointment
they expect the customer to have. This will be fed to the summary document
generator so TAP then becomes the canonical source for this piece of
information.

screen shot 2017-03-31 at 12 50 26

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -130,7 +130,8 @@ class AppointmentsController < ApplicationController
       :memorable_word,
       :notes,
       :opt_out_of_market_research,
-      :status
+      :status,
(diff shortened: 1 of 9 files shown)


benlovell:
[As a guider I want the option to select if the appointment is a 50-54 or 55+ type of appointment](https://trello.com/c/hsR0bRze/90-as-a-guider-i-want-the-option-to-select-if-the-appointment-is-a-50-54-or-55-type-of-appointment)

benlovell:
Deployment should be coordinated with guidance-guarantee-programme/output#280.

Guntrisoft:
👍

PR #503, opened 2017-04-03
Bump default start date for calendars by an hour

benlovell:
This will account for timezone discrepancies and is another ugly hack
(tm) however temporary, to workaround the timezone issues until we get
a proper workable solution in place.

Resource managers were seeing slots that have actually passed due to
the discrepancy between UTC and BST. This change ensures we add another
hour grace period to the first potentially available slot.

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -15,7 +15,7 @@ class BookableSlot < ApplicationRecord
   end
 
   def self.next_valid_start_date(user = nil)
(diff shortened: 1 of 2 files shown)


benlovell:
[Selected time slot is not re-highlighted when there is an error on creating a new appointment](https://trello.com/c/H7CQHZBr/94-selected-time-slot-is-not-re-highlighted-when-there-is-an-error-on-creating-a-new-appointment)

PR #504, opened 2017-04-04
Reselect chosen slot after roundtripping

benlovell:
This ensures we parse the dates as UTC before comparison and thus
causes the originally selected slot to be highlighted after the form
roundtrips.


benlovell:
[Selected time slot is not re-highlighted when there is an error on creating a new appointment](https://trello.com/c/H7CQHZBr/94-selected-time-slot-is-not-re-highlighted-when-there-is-an-error-on-creating-a-new-appointment)

Guntrisoft:
👍

PR #507, opened 2017-04-05
Update appointment type from date of birth

benlovell:
This change preselects the correct appointment type from the provided date
of birth when booking / modifying an appointment. It's possible the agent,
guider or resource manager will want to override this if they spot the
appointment is within a few months of a particular range - so this is also
possible.

### DIFF
--- a/app/assets/javascripts/modules/customer-age.es6
+++ b/app/assets/javascripts/modules/customer-age.es6
@@ -45,6 +45,8 @@
       age = Math.floor(today.diff(inputDate, 'year'));
 
       if (age) {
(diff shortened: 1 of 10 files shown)


benlovell:
[As a guider I want the option to select if the appointment is a 50-54 or 55+ type of appointment](https://trello.com/c/hsR0bRze/90-as-a-guider-i-want-the-option-to-select-if-the-appointment-is-a-50-54-or-55-type-of-appointment)

Guntrisoft:
Looks good to me 👍

PR #508, opened 2017-04-06
Fix timezone discrepancy for past appointments

benlovell:
The `moment` or `now` is local time but the appointment end is UTC.
A comparison was made to highlight past appointments during the
course of the day.

This resolves the discrepancy (using an ugly hack) until we can put
proper timezone handling in place. We clone the end date so we don't
mutate the actual underlying time.


benlovell:
[Past appointments should be highlighted correctly according to time](https://trello.com/c/nUlBU3xQ/107-past-appointments-should-be-highlighted-correctly-according-to-time)

PR #510, opened 2017-04-11
Ensure notifications are directed accurately

benlovell:
In the case of dropped summary documents these are always the guider's
responsibility to action, so only they will receive the high-priority
notifications.

In the case of other dropped emails - for now we will notify both the
guider and the agent - so long as the agent is not the Pension Wise website
API agent. This is despite the `owner` of the activity always being the
guider.

Since these two activities are always considered 'high priority' we can do
away with the guard and move the event hooks into the concrete activity
classes.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -64,7 +64,6 @@ class Activity < ApplicationRecord
 
   after_commit if: :high_priority? do
     PusherHighPriorityCountChangedJob.perform_later(owner)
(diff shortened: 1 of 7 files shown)


benlovell:
[As a guider I am notified when a summary document email fails/drops](https://trello.com/c/iOmhgoTk/111-as-a-guider-i-am-notified-when-a-summary-document-email-fails-drops)

benlovell:

Guntrisoft:
:+1:

PR #511, opened 2017-04-12
Notify individuals based on the type of message

benlovell:
We now notify and assign responsibility for dropped / bounced emails to the
agent making the booking. For any other dropped emails we assign the
responsibility and notify the guider.

Mailgun returns the original headers we supply when sending the emails so
we pick the `message_type` reliably from the inbound webhook.


benlovell:

PR #512, opened 2017-04-13
Resolve outstanding high-priority activities

benlovell:
Now we correctly assign responsibility for these it makes sense to mark
the existing activities as resolved so the agents and guiders have a
clean break.


benlovell:

Guntrisoft:
You da best 👍

PR #513, opened 2017-04-13
Count public holidays if calculating working days

benlovell:
We were not previously counting bank holidays or other public holidays
when advancing by 'working days' so was a chance customers could
actually book within the lock down period prior to this change.


benlovell:
[Ensure holidays are configured in 'working days'](https://trello.com/c/WZpOfeaG/123-ensure-holidays-are-configured-in-working-days)

PR #515, opened 2017-04-25
Show schedule while on holiday

Guntrisoft:

### DIFF
--- a/app/assets/javascripts/modules/calendar.es6
+++ b/app/assets/javascripts/modules/calendar.es6
@@ -164,13 +164,18 @@ class Calendar extends TapBase {
   }
 
   eventRender(event, element) {
-    const now = moment();
+    const now = moment(),
+      elementId = `event_${event._id}`;
     let end = event.end.clone();
 
     // account for timezone weirdness
     if (end.subtract(1, 'h') < now) {
       element.addClass('fc--past');
     }
+
+    $(element).attr('id', elementId);
+
+    event.elementId = elementId;
   }
 
   eventAfterRender() {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/modules/calendar.es6
```diff
@@ -164,13 +164,18 @@ class Calendar extends TapBase {
...
     let end = event.end.clone();
 
     // account for timezone weirdness
     if (end.subtract(1, 'h') < now) {
       element.addClass('fc--past');
     }
+
+    $(element).attr('id', elementId);
+
+    event.elementId = elementId;
```
Guntrisoft: Storing an ID reference to the DOM element in the event object.

This is instead of storing the jQuery collection object (see appointment-availability.es6 below) which was causing issues with circular references in the tests.

Guntrisoft:
[As a resource manager I want to see the bookable slots when people are blocked out on holiday](https://trello.com/c/j4dD0WGj/118-as-a-resource-manager-i-want-to-see-the-bookable-slots-when-people-are-blocked-out-on-holiday)

app/controllers/bookable_slots_controller.rb
```diff
@@ -21,10 +21,9 @@ def scoped_to_me?
 
   def bookable_slots
     slots = BookableSlot
-            .without_holidays
             .within_date_range(start_at, end_at)
 
-    slots = slots.for_guider(current_user) if scoped_to_me?
+    slots = slots.without_holidays.for_guider(current_user) if scoped_to_me?
```
benlovell: The commit references the wrong URI for this resource.

PR #521, opened 2017-04-27
Rework `Appointment.needing_reminder`

jongilbraith:
This should now:

- observe a 48 hour window ahead of appointments for sending reminders
- not resend reminders when called multiple times during that window
- once 48 hours from the first reminder passes (i.e. approximately original time of the appointment when first reminded), will start sending reminders again should we re-encounter the 48 hour window ahead of an appointment (i.e. if it's been rescheduled into the future)

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -36,15 +36,19 @@ class Appointment < ApplicationRecord
   scope :not_cancelled, -> { where.not(status: %i(cancelled_by_customer cancelled_by_pension_wise)) }
 
   def self.needing_reminder
+    window = Time.zone.now..48.hours.from_now.in_time_zone
+
     pending
-      .includes(:activities)
-      .where(start_at: Time.zone.now..24.hours.from_now)
-      .where.not(
-        email: '',
-        activities: {
-          type: 'ReminderActivity', created_at: 12.hours.ago..12.hours.from_now
-        }
-      )
+      .where(start_at: window)
+      .where.not(email: '', id: reminded_ids)
+  end
+
+  def self.reminded_ids
+    window = 48.hours.ago.in_time_zone..Time.zone.now
+
+    ReminderActivity
+      .where(created_at: window)
+      .pluck(:appointment_id)
   end
 
   validates :agent, presence: true
--- a/spec/models/appointment_spec.rb
+++ b/spec/models/appointment_spec.rb
@@ -399,52 +399,76 @@ RSpec.describe Appointment, type: :model do
   end
 
   describe '.needing_reminder' do
-    let(:appointment) do
-      create(
-        :appointment,
-        start_at: BusinessDays.from_now(10)
-      )
-    end
+    context 'more than 48 hours before the appointment' do
+      it 'does not return the appointment' do
+        appointment = create(:appointment, start_at: BusinessDays.from_now(10))
 
-    context 'more than 24 hours until appointment' do
-      it 'does not need a reminder' do
         expect(Appointment.needing_reminder).to_not include(appointment)
       end
     end
 
-    context 'less than 24 hours until appointment' do
-      let(:result) do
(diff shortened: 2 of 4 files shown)


app/models/reminder_activity.rb
```diff
@@ -1,2 +1,6 @@
 class ReminderActivity < Activity
+  def self.already_reminded_appointment_ids
```
jongilbraith: Not 100% on whether this would be more at home in `Appointment`.

Originally was just a local var in the `Appointment.needing_reminder` method but got my wrist slapped by Rubocop for that.

jongilbraith:
[As a customer who has an appointment booked I receive the appointment reminder email in adequate time so that I can prepare for my appointment](https://trello.com/c/Akr7hqWV/136-as-a-customer-who-has-an-appointment-booked-i-receive-the-appointment-reminder-email-in-adequate-time-so-that-i-can-prepare-for-)

app/models/reminder_activity.rb
benlovell: Seems legit. Though perhaps the `already_` can be dropped in favour of just `reminded_appointment_ids` as the 'already' part is implied by 'reminded'.

app/models/reminder_activity.rb
```diff
@@ -1,2 +1,6 @@
...
+    where(created_at: (Time.zone.now - 48.hours)..Time.zone.now)
```
benlovell: Do you reckon `where(created_at: 48.hours.ago.in_time_zone..Time.zone.now)` is any more concise?

spec/models/appointment_spec.rb
```diff
@@ -437,14 +437,40 @@
         end
       end
 
-      context 'appointment has been rescheduled' do
-        it 'needs another reminder' do
-          appointment.update!(
-            start_at: BusinessDays.from_now(20),
-            end_at:   BusinessDays.from_now(20) + 1.hour
-          )
+      context 'has already had a reminder' do
```
benlovell: I wonder whether those extra scenarios in the feature spec you added might be better expressed here instead... Given what they're ultimately testing is the inputs / outputs to the `needing_reminder` scope. Anyway, not overly concerned about this since the features are quick enough to run, but it's worth keeping an eye on if the logic grows any further perhaps.

app/models/appointment.rb
```diff
@@ -37,14 +37,8 @@ class Appointment < ApplicationRecord
...
     pending
-      .includes(:activities)
-      .where(start_at: Time.zone.now..24.hours.from_now)
-      .where.not(
-        email: '',
-        activities: {
-          type: 'ReminderActivity', created_at: 12.hours.ago..12.hours.from_now
-        }
-      )
+      .where(start_at: Time.zone.now..(Time.zone.now + 48.hours))
```
benlovell: Same as below but with the `Time.zone.now..48.hours.from_now.in_time_zone` perhaps more concise, though I'm not totally sure...

jongilbraith: Yeah, tbh my first draft just had it all in a single method with local vars of `now`, `two_days_ago` and `two_days_from_now`.

Rubocop didn't like it but I thought it was decent.

Then after breaking it up into two methods I didn't like assigning the local vars that only got called the one time so went back to what we have now.

app/models/reminder_activity.rb
jongilbraith: Ah, yes, `in_time_zone` seems nice. Like that.

spec/models/appointment_spec.rb
jongilbraith: I'll revisit this. I felt the same way but wanted to get it out the door, but it's the right thing to do.

benlovell: I admire your commitment 👍 💪

PR #522, opened 2017-05-01
Switch to `phantomjs` gem for binaries

benlovell:
For some reason (probably `exec`ing a new process every time, but I
didn't check) the binaries and process overhead from
`phantomjs-binaries` adds around an extra 30 seconds to an `rspec` run
on my machine. We switched away from `phantomjs` in the past since it
used an unreliable bitbucket CDN that was often rate-limited in Travis,
it seems this has been addressed now so switching back to this gem
yields a ~30 second reduction in spec runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ services:
 cache:
   bundler: true
   directories:
+  - $HOME/.phantomjs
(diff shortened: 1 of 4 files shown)


Guntrisoft:
👍

PR #526, opened 2017-05-15
Calculate type from age at appointment

benlovell:
This ensures the appointment type is correctly inferred from the customers age
at the time of their appointment.


benlovell:
[The type of appointment should be calculated from the age of the customer when the appointment takes place](https://trello.com/c/4gG5MLUQ/168-the-type-of-appointment-should-be-calculated-from-the-age-of-the-customer-when-the-appointment-takes-place)

PR #530, opened 2017-06-16
Display time zone information in customer emails

jongilbraith:
**Example:**

screen shot 2017-06-16 at 15 03 31


jongilbraith:
[As a customer who has booked an appointment abroad I am aware that the appointment time is in BST.](https://trello.com/c/qG2IqNKI/197-as-a-customer-who-has-booked-an-appointment-abroad-i-am-aware-that-the-appointment-time-is-in-bst)

PR #531, opened 2017-06-20
Thoroughly guard against appointment activities

benlovell:
We need to thoroughly guard against notification activities for appointments
without customer-supplied email addresses. There have been ~500 created for
appointments and this bug was reported by TPAS.

Review commit-by-commit.


benlovell:
[Ensure email activities are only created when emails are sent](https://trello.com/c/BDhOxrIY/208-ensure-email-activities-are-only-created-when-emails-are-sent)

PR #532, opened 2017-06-26
Prevent hiding of bookable slots that are adjacent to holidays

jongilbraith:
So, the solution turned out to be as simple as what I could swear I tried right at the start.

Possibly one red herring was that the test data used ActiveSupport's `end_of_day` time helper which set the end of the seeded holidays to 23:59:59 whereas holidays created within tap are 10:00 - 11:00, etc, which threw my initial assessment of the situation.

I've included a change to the seeds to address this.

Some screencaps of the behaviour below.

Holiday just before a bookable slot

Holiday just after a bookable slot

Holiday overlapping a bookable slot

### DIFF
--- a/app/assets/javascripts/modules/calendars/company.es6
+++ b/app/assets/javascripts/modules/calendars/company.es6
@@ -243,8 +243,8 @@ class CompanyCalendar extends Calendar {
           )
           &&
           (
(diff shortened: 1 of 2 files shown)


jongilbraith:
[As resource manager I am able to allocate an appointment when a blocked out slot ends](https://trello.com/c/v3GH4By2/213-as-resource-manager-i-am-able-to-allocate-an-appointment-when-a-blocked-out-slot-ends)

PR #533, opened 2017-07-06
Bump rails to latest version

jongilbraith:
Updated rails version constraint and `bundle update`

Mostly minor bumps of gems, with a few exceptions:

* `govuk_admin_template` gone from [6.1.0 to 6.2.0](https://github.com/alphagov/govuk_admin_template/blob/master/CHANGELOG.md)
* `oj` from [3.1.3 to 3.3.0](https://github.com/ohler55/oj/blob/master/CHANGELOG.md)

Nothing that looks like needing addressing.


Gemfile
```diff
@@ -38,7 +38,7 @@ source 'https://rubygems.org' do
   gem 'postgres-copy'
   gem 'puma', '~> 3.0'
   gem 'pusher'
-  gem 'rails', '~> 5.1.1'
+  gem 'rails', '~> 5.1.2'
```
benlovell: This isn't strictly necessary. Can we leave it as-is and let `bundle update rails` do its thing?

PR #534, opened 2017-07-19
[WIP] Assign regular drops to TP agents

benlovell:
Previously, we were assigning high-priority activities to the agent for
the appointment created notification and guiders for all other message
types. TPAS have requested that their guiders only receive dropped
summary document notifications so this change ensures we abide by this.

### DIFF
--- a/app/forms/drop_form.rb
+++ b/app/forms/drop_form.rb
@@ -30,7 +30,6 @@ class DropForm
     DropActivity.from(
       event,
       description,
-      message_type,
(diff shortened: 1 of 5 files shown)


benlovell:
[As a phone guider I want to be notified only when a digital summary document email fails/drops](https://trello.com/c/y3Wz64eS/228-as-a-phone-guider-i-want-to-be-notified-only-when-a-digital-summary-document-email-fails-drops)

PR #537, opened 2017-07-28
Allow appointments up to 40 days in advance

benlovell:
We were previously locked to 30 working days, though the booking window
elsewhere was increased to 8 weeks.


benlovell:
[Ensure appointments can be placed up to 40 working days in the future](https://trello.com/c/nF6BjnaK/250-ensure-appointments-can-be-placed-up-to-40-working-days-in-the-future)

PR #545, opened 2017-10-18
Revert changes required to accomodate TZ change

benlovell:
This is hacky as hell but once merged will revert to the GMT timezone
correctly.

When I get a chance I will add proper timezone support. I promise.


benlovell:
[Daylight saving time changes](https://trello.com/c/18nmIzwP/130-daylight-saving-time-changes)

PR #547, opened 2017-11-09
Don't validate presence of WDYH when rebooked

benlovell:
Some bookings prior to the WDYH change will be unspecified so when
these are the basis of a rebooking we ought not to validate for WDYH
presence.


benlovell:
[As a TPAS guider I am able to rebook an appointment that has 'where did you hear about us' unspecified](https://trello.com/c/KrI3iU1l/335-as-a-tpas-guider-i-am-able-to-rebook-an-appointment-that-has-where-did-you-hear-about-us-unspecified)

PR #549, opened 2017-11-29
Send SMS reminders 48hr before appointments

benlovell:
Sends SMS reminders via Notify, 48 hours before an appointment is due.
This needs to be scheduled.


benlovell:
[As a customer who has booked a phone appointment I receive a reminder SMS](https://trello.com/c/cdxYiyun/334-as-a-customer-who-has-booked-a-phone-appointment-i-receive-a-reminder-sms)

PR #552, opened 2017-12-13
Create recurring holidays

benlovell:
When creating single day holidays with an associated start and end
time, the resource manager now has the option to specify recurrence.
When they provide a recurrence date the holiday is repeated for each
weekday between the start date and the provided end date. This behaves
similarly to all-day holidays but with the option to have shorter
periods for each day.


benlovell:
[As a resource manager I am able to specify times of when I want to block out multiple days](https://trello.com/c/9a6MRVpH/338-as-a-resource-manager-i-am-able-to-specify-times-of-when-i-want-to-block-out-multiple-days)

PR #553, opened 2017-12-18
Persist last date via calendar cookie

benlovell:
This was not enabled for this particular calendar but TPAS have
requested we change this behaviour.


benlovell:
[As a resource manager I want the allocations calendar to remember my previously selected day](https://trello.com/c/JvruYqxu/361-as-a-resource-manager-i-want-the-allocations-calendar-to-remember-my-previously-selected-day)

PR #557, opened 2017-12-27
Allow resource managers to adhoc reschedule

benlovell:
Grants resource managers the ability to reschedule appointments on an
adhoc basis, ie not from the calendar schedule. When rescheduling the
resource manager is presented with a further option to select from
adhoc availability. When they select this they are then presented with
a date and guider to choose from.


benlovell:
[As a resource manager I want to create an appointment directly in TAP without a bookable slot](https://trello.com/c/yMm0AGIz/341-as-a-resource-manager-i-want-to-create-an-appointment-directly-in-tap-without-a-bookable-slot)

PR #558, opened 2018-01-03
Correctly disable weekends

benlovell:
This was previously misconfigured.


benlovell:
[As a resource manager I want to create an appointment directly in TAP without a bookable slot](https://trello.com/c/yMm0AGIz/341-as-a-resource-manager-i-want-to-create-an-appointment-directly-in-tap-without-a-bookable-slot)

PR #560, opened 2018-01-09
Swallow and publish notify client errors

benlovell:
The only failures we see are related to invalid numbers being provided
by our service. There is no need for these to be retried yet should be
captured for operational metrics.


benlovell:
[Gracefully handle notify SMS rejections](https://trello.com/c/yoApcHBX/372-gracefully-handle-notify-sms-rejections)

PR #561, opened 2018-01-16
Do not send reminders for fresh appointments

benlovell:
Now appointments can be created at very short notice we need to guard
against this particular circumstance. By extending the window a few
hours we allow a grace period before sending a reminder.


benlovell:
[Ensure reminders are not sent for appointments occurring on the same day](https://trello.com/c/qkZyCanZ/377-ensure-reminders-are-not-sent-for-appointments-occurring-on-the-same-day)

PR #567, opened 2018-02-15
Explicitly wait for the field to be cleared

benlovell:
This was causing intermittent timing issues during spec runs in the CI
environment.


benlovell:
[Fix the timing issue during message activity feature specs](https://trello.com/c/9xQ728sJ/408-fix-the-timing-issue-during-message-activity-feature-specs)

PR #569, opened 2018-03-13
Accomodate BST timezone

benlovell:
Ensures we can accomodate the BST timezone change due later this month.

This should be merged and deployed by 1AM 25/03/18.


benlovell:
[Prepare for timezone change to DST](https://trello.com/c/6QYT4ihO/438-prepare-for-timezone-change-to-dst)

PR #573, opened 2018-03-23
Extend permissable booking window

benlovell:
This resolves the discrepancy between the slots being displayed in the
internal booking journey and the latest slot actually allowed.


benlovell:
[Allow appointments up to the end of the displayed window](https://trello.com/c/J6TvcaOB/450-allow-appointments-up-to-the-end-of-the-displayed-window)

PR #575, opened 2018-03-28
Ensure validation errors on `start_at` display

benlovell:
These were not being shown when present and this causes confusion to
the agents when attempting to place appointments.


benlovell:
[Display errors if present on the appointment start date and time](https://trello.com/c/duZJuCL2/451-display-errors-if-present-on-the-appointment-start-date-and-time)

PR #576, opened 2018-03-28
Do not alert guiders on drop notifications

benlovell:
This ensures guiders are not alerted or assigned high-priority
activities when an email is permanently failed by mailgun. They will
continue to receive notifications and assignments for dropped summary
documents.


benlovell:
[As a TPAS guider, I do not need to see email delivery notification failures from the book telephone appointment stage](https://trello.com/c/cmSz9K85/455-as-a-tpas-guider-i-do-not-need-to-see-email-delivery-notification-failures-from-the-book-telephone-appointment-stage)

PR #577, opened 2018-04-09
Associate high priority activities with owner

benlovell:
These were incorrectly determined by the associated appointment's
guider or agent. The `owner` is the person directly responsible for the
resolution of the activity so it should only show up in their
feed/alerts.


benlovell:
[Investigate issues with high-priority activities and assignment](https://trello.com/c/3qzHnCiM/478-investigate-issues-with-high-priority-activities-and-assignment)

PR #578, opened 2018-04-10
Configure 2018 bank holidays

benlovell:
This ensures the next working day logic is calculated correctly when
booking appointments or displaying availability.


benlovell:
[Configure TAP bank holidays for calculation of 'working days'](https://trello.com/c/TEr1iNFP/479-configure-tap-bank-holidays-for-calculation-of-working-days)

PR #579, opened 2018-04-12
Reduce grace period to one business day

benlovell:
Cuts the initial grace period for bookings down to one business day, as
requested by TPAS.


benlovell:
[As a Pension Wise customer and/or TP agent, I can see bookable slots which are available after end of next working day](https://trello.com/c/frzmxdB2/462-as-a-pension-wise-customer-and-or-tp-agent-i-can-see-bookable-slots-which-are-available-after-end-of-next-working-day)

PR #581, opened 2018-04-16
Validate only an email/address is present

benlovell:
Don't permit both an email address and a postal address for
confirmation as they should be mutually exclusive.


benlovell:

PR #582, opened 2018-04-20
Tag the `email` field with the errors

benlovell:
These errors were previously attributed to `base` and weren't being
displayed sensibly. Tagging them to the `email` field ensures they
appear in the correct context when creating an appointment.


benlovell:

PR #583, opened 2018-04-23
Summarise availability data for reporting

benlovell:
This will be scheduled to run every morning before the day's
appointments are due to start. It summarises availability over the
following four week period, inside of the regular booking window.

At the moment this is only concerned with TPAS guider appointments but
once we support multiple providers it will address other provider's
appointments also.


benlovell:
[As a research or business analyst, I require access to reporting which indicates the 'availability' of appointments](https://trello.com/c/GnF4FDHV/393-as-a-research-or-business-analyst-i-require-access-to-reporting-which-indicates-the-availability-of-appointments)

PR #584, opened 2018-04-25
Store status transitions upon creation or changes

benlovell:
Stores the status transition in an association when it is initially created
or upon a status change.


benlovell:
[As an analyst, I need to be able to view and analyse a telephone customer's journey and transitioning appointment statuses](https://trello.com/c/s7ucTMfN/498-as-an-analyst-i-need-to-be-able-to-view-and-analyse-a-telephone-customers-journey-and-transitioning-appointment-statuses)

PR #585, opened 2018-04-29
Allow customers to cancel by SMS

benlovell:
This permits customers to cancel by replying to their appointment reminder
SMS. Upon successful cancellation the following events occur:

* The status is changed to `cancelled_by_customer_sms`
* An activity entry is created for the audit trail
* The associated guider is notified, if the appointment occurs today
* The customer receives an SMS reply, in confirmation


benlovell:
[As a customer who has booked a phone appointment I can cancel an appointment by replying to SMS reminder](https://trello.com/c/dlV3keN4/402-as-a-customer-who-has-booked-a-phone-appointment-i-can-cancel-an-appointment-by-replying-to-sms-reminder)

PR #587, opened 2018-05-02
Include SMS cancellations with regular ones

benlovell:
Since the introduction of SMS cancellations we should include the newly
introduced `cancelled_by_customer_sms` status when querying by the
previous cancellation statuses to ensure reports are accurate.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -4,6 +4,12 @@ class Appointment < ApplicationRecord
 
   acts_as_copy_target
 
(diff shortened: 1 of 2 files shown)


benlovell:
[As a customer who has booked a phone appointment I can cancel an appointment by replying to SMS reminder](https://trello.com/c/dlV3keN4/402-as-a-customer-who-has-booked-a-phone-appointment-i-can-cancel-an-appointment-by-replying-to-sms-reminder)

PR #589, opened 2018-05-16
Summarise availability per provider

benlovell:
The currently released summary is generated based on TPAS availability only
for the timebeing. This change ensures availability is correctly
summarised, per organisation.


benlovell:
[TAP reporting summary produces results across multiple providers](https://trello.com/c/KwWM2Vw0/493-tap-reporting-summary-produces-results-across-multiple-providers)

PR #591, opened 2018-05-31
appointments need sms reminders at both 7 days and 2 days

spannerinworks:
Increase the window for sms reminders to be both 7 days before and 2 days before.

https://trello.com/c/uLUWENbE

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -224,10 +224,8 @@ class Appointment < ApplicationRecord
   end
 
   def self.needing_sms_reminder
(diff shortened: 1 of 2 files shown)


benlovell:
[As a customer who has booked a telephone appointment I receive an earlier SMS message, 7 days prior to the appointment.](https://trello.com/c/uLUWENbE/541-as-a-customer-who-has-booked-a-telephone-appointment-i-receive-an-earlier-sms-message-7-days-prior-to-the-appointment)

PR #593, opened 2018-06-11
Do not send reminders on same day as booked

benlovell:
We are sending reminders immediately after confirmations are sent in
the case of appointments booked that immediately fall within the 2 or 7
day reminder period for SMS and email reminders. This change ensures we
do not send these reminders unnecessarily.


benlovell:
[Ensure that telephone appointment customers do not receive reminder if booking within 2 or 7 day window](https://trello.com/c/4NZA26NG/572-ensure-that-telephone-appointment-customers-do-not-receive-reminder-if-booking-within-2-or-7-day-window)

PR #594, opened 2018-06-17
Switch CAS organisation ID

benlovell:
This has been shuffled around in signon to match Drumpchapel's previous
content ID, this is to allow booking managers from CAS/Drumchapel to
use both the planner and TAP in unison.


benlovell:
[Align CAS and Drumchapel's organisation identifiers](https://trello.com/c/EWDKbG8c/581-align-cas-and-drumchapels-organisation-identifiers)

PR #597, opened 2018-06-26
Add some determinism to guider selection

benlovell:
Prior to this change there were instances of guiders being allocated
across the organisation boundaries when they had identically available
slots. This change ensures only TP agents can allocate across the
organisation boundaries. All other permeations of role/organisation are
restricted to their own guiders.


benlovell:
[Ensure TPAS, TP & CAS Resource Managers are unable to schedule appointments to a slot not offered by their own organisation](https://trello.com/c/KSCpvnlc/586-ensure-tpas-tp-cas-resource-managers-are-unable-to-schedule-appointments-to-a-slot-not-offered-by-their-own-organisation)

PR #599, opened 2018-06-28
Don't process print confirmations on day created

benlovell:
Ensure we don't needlessly process a print confirmation on the same day
as the appointment.


benlovell:
[Print confirmations should not be processed on the same day appointments take place](https://trello.com/c/thIYx1Te/592-print-confirmations-should-not-be-processed-on-the-same-day-appointments-take-place)

PR #600, opened 2018-07-02
Include timezone with SMS reminder

benlovell:
We do this elsewhere and it's important to include for telephone
customers as they can be situated anywhere in the world.


benlovell:
[Add timezone to SMS reminder](https://trello.com/c/T7xKyJFs/596-add-timezone-to-sms-reminder)

PR #602, opened 2018-07-05
Display missing owner as 'The system'

benlovell:
There are some edge-cases where the auditing 'owner' will be missing. In
these instances we can assume the change was made by 'The system' and
display this accordingly.


benlovell:
[Audits without an 'owner' raise an error when viewed for a particular appointment](https://trello.com/c/mjgDTsdk/585-audits-without-an-owner-raise-an-error-when-viewed-for-a-particular-appointment)

PR #603, opened 2018-07-12
Allow TPAS managers to reschedule during conf

benlovell:
Allows TPAS resource managers to reschedule appointments during guider
conference days that would otherwise be blocked out.

I had to also fix a bunch of stupid time-sensitive specs.


benlovell:
[Lift exclusion dates for 17 & 18 July (TPAS)](https://trello.com/c/BEhyMwDn/614-lift-exclusion-dates-for-17-18-july-tpas)

PR #605, opened 2018-08-07
Remove guider conference checks

benlovell:
These are no longer necessary so can be removed.


benlovell:
[Remove guider conference specific rules and validation](https://trello.com/c/QlKarCnO/638-remove-guider-conference-specific-rules-and-validation)

PR #607, opened 2018-08-21
Display the guider's organisation

benlovell:
Displays the organisation in the search index results.


benlovell:
[As a call centre agent I need to be able to see which Delivery Partner telephony guiders works for on TAP after the appointment is booked.](https://trello.com/c/QjkNqyds/656-as-a-call-centre-agent-i-need-to-be-able-to-see-which-delivery-partner-telephony-guiders-works-for-on-tap-after-the-appointment)

PR #610, opened 2018-10-03
Move slot picker below customer details

benlovell:
Moves the date picker under the customer details form. This is to
mitigate slots being taken during the time it takes for the agent to
complete the customer details.


benlovell:
[Move slot picker to bottom of TAP agent booking form with drop down to update in real time](https://trello.com/c/GtKmy1Zj/688-move-slot-picker-to-bottom-of-tap-agent-booking-form-with-drop-down-to-update-in-real-time)

PR #611, opened 2018-10-16
Allow for timezone change

benlovell:
Super hacky hack to handle the upcoming timezone change.


benlovell:
[Address DST/GMT/UTC issues on TAP](https://trello.com/c/u6WEH7WT/678-address-dst-gmt-utc-issues-on-tap)

PR #613, opened 2018-10-27
Copy `default` configuration to production

benlovell:
This ensures the correct pool size is configured in the production
environment.


benlovell:
[Configure database connection pool size](https://trello.com/c/tn1Wo66X/730-configure-database-connection-pool-size)

PR #614, opened 2018-11-07
Use 'business days' logic to calculate window

benlovell:
Previously the end of the booking window was calculated at 8 weeks from the
time of request and didn't consider bank holidays. This change ensures we
also skip those days when calculating the end of the window.


benlovell:
[Configure F2F Planner & TAP so the booking window used the same library/logic.](https://trello.com/c/LVpP9vCr/737-configure-f2f-planner-tap-so-the-booking-window-used-the-same-library-logic)

PR #615, opened 2018-11-07
Reduce starting window to 1 working day

benlovell:


benlovell:
[As a Pension Wise customer and/or TP agent, I can see bookable slots on TAP which are available after end of next working day](https://trello.com/c/w617ahiB/736-as-a-pension-wise-customer-and-or-tp-agent-i-can-see-bookable-slots-on-tap-which-are-available-after-end-of-next-working-day)

PR #617, opened 2019-01-21
Ensure appointment organisations cannot change

benlovell:
Appointments should never switch organisations. This validation ensures
when the guider changes, they can only ever originate from the same
provider/organisation.


benlovell:
[Enforce validation so that TP agents and agent managers are unable to reschedule / book appointments at short notice or reschedule across DP organisations.](https://trello.com/c/KdomAS2w/842-enforce-validation-so-that-tp-agents-and-agent-managers-are-unable-to-reschedule-book-appointments-at-short-notice-or-reschedule)

PR #620, opened 2019-02-12
Ensure grace period is respected

benlovell:
Resource managers should see slots within the grace period but the
recent changes for segmentation caused a regression. This change ensures
resource managers can once again book short notice appointments.


benlovell:
[Ensure resource manager grace period is respected](https://trello.com/c/1WP62nc7/887-ensure-resource-manager-grace-period-is-respected)

PR #623, opened 2019-03-07
Add NI as a telephony provider

benlovell:
NI will be offering telephony appointments so they'll need to be
configured as a provider.


benlovell:
[Add 'Northern Ireland' to TAP (Other Providers)](https://trello.com/c/yuXKIhpg/902-add-northern-ireland-to-tap-other-providers)

PR #625, opened 2019-03-19
Leave historical slots when generating

benlovell:
Previously the slots would be deleted in the past when a guider's
schedule was generated. This stops us being able to report on
availability retrospectively. This change ensures we only delete future
slots, and only in the case of guider deactivation.

### DIFF
--- a/app/models/bookable_slot.rb
+++ b/app/models/bookable_slot.rb
@@ -135,9 +135,7 @@ class BookableSlot < ApplicationRecord
   end
 
   def self.generate_for_guider(guider)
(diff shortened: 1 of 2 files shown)


benlovell:
[As an analyst, I need to know how many bookable slots are free in TAP and Planner to ensure correct resourcing is in place](https://trello.com/c/Dsmpe4bK/871-as-an-analyst-i-need-to-know-how-many-bookable-slots-are-free-in-tap-and-planner-to-ensure-correct-resourcing-is-in-place)

PR #628, opened 2019-03-29
Revert "Allow for timezone change"

benlovell:
Super hacky-hack time to handle the timezone change this weekend.

This reverts commit e948e168b78db45f8e6757bc7067fcc83591ad80.


benlovell:
[As a customer who has booked an appointment I need to be aware that an appointment time is in DST/BST](https://trello.com/c/C10aCr9E/930-as-a-customer-who-has-booked-an-appointment-i-need-to-be-aware-that-an-appointment-time-is-in-dst-bst)

PR #629, opened 2019-04-10
Allow CAS to take bookings during Easter

benlovell:
Some CAS guiders will be available during this period so these changes
ensure only CAS guiders (without explicit holiday blocks) are made
available during the Easter bank holiday period.

This will need to be ran on production and the usual, nightly bank
holiday job disabled to ensure new bank holidays for this period are not
created.


benlovell:
[Remove specific Bank Holiday restrictions for  Drumchapel Planner and CAS TAP](https://trello.com/c/dIDBHg6S/885-remove-specific-bank-holiday-restrictions-for-drumchapel-planner-and-cas-tap)

PR #630, opened 2019-04-18
Allow May bank holiday bookings for CAS

benlovell:


benlovell:
[Remove May Bank Holiday restrictions for CAS TAP and Drumchapel Planner](https://trello.com/c/2PqkKmMn/967-remove-may-bank-holiday-restrictions-for-cas-tap-and-drumchapel-planner)

PR #631, opened 2019-05-01
Tag logs with the current user ID

benlovell:
This should make ownership easier to determine when something untoward
happens.


benlovell:
[Investigate issues with NI guiders and availability](https://trello.com/c/YgF17SMp/978-investigate-issues-with-ni-guiders-and-availability)

PR #632, opened 2019-05-07
Redact customer details on request

benlovell:
Redacts a customer's personal details and removes any associated audits
or activities which would also expose any personal information.


benlovell:
[Remove booking/customer details from TAP #2](https://trello.com/c/vLa33F37/979-remove-booking-customer-details-from-tap-2)

PR #633, opened 2019-06-14
Include date in SMS cancellation

benlovell:
Includes the date attribute so it can be embedded into the SMS
cancellation success template.


benlovell:
[Update sms reminder messages on Notify](https://trello.com/c/CowMqGmw/1023-update-sms-reminder-messages-on-notify)

PR #635, opened 2019-06-18
Add CITA Wallsend provider

benlovell:
Adds the new CITA Wallsend provider, they have the same general
permissions as the other providers.


benlovell:
[Add Citizens Advice to TAP](https://trello.com/c/e4gf2N1K/1014-add-citizens-advice-to-tap)

PR #637, opened 2019-07-19
Enforce edit permission

benlovell:
This wasn't enforced explicitly before so when viewing duplicates,
sometimes a cross-organisation appointment would be linked and therefore
editable.


benlovell:
[Investigate duplicate booking for TPAS/CAS and visibility permissions](https://trello.com/c/lSrqP6OD/1073-investigate-duplicate-booking-for-tpas-cas-and-visibility-permissions)

PR #642, opened 2019-09-24
Send accessibility alerts to resource managers

benlovell:
When a customer or agent specifies that they need an accessibility
adjustment to access their appointment we should notify those resource
managers associated with the appointment ahead of time.


benlovell:
[As a resource manager I should receive an email notification where a customer indicates they need help or an adjustment to access their Pension Wise appointment](https://trello.com/c/PDhzssmP/1127-as-a-resource-manager-i-should-receive-an-email-notification-where-a-customer-indicates-they-need-help-or-an-adjustment-to-acces)

PR #647, opened 2019-10-03
Alert supervisors mailbox for TPAS guiders

benlovell:
When the guider is from TPAS and we send an accessibility adjustment
alert we must send them to the supervisors mailbox, otherwise they're
sent to the resource managers as usual.


benlovell:
[As a TPAS Resource Manager, I should not receive accessibility alert emails as these are better dealt with by Shift Supervisors](https://trello.com/c/5wBBIlC0/1143-as-a-tpas-resource-manager-i-should-not-receive-accessibility-alert-emails-as-these-are-better-dealt-with-by-shift-supervisors)

PR #648, opened 2019-10-14
Modifies the scheduling templates

benlovell:
This renames existing templates and adds a new template for the 'late'
shift as requested by TPAS for their seasonal guider allocations.


benlovell:
[As a TPAS resource manager, I would like to make changes to each guider’s TAP schedule so that they are based on existing operational requirements](https://trello.com/c/mfwcItcC/1153-as-a-tpas-resource-manager-i-would-like-to-make-changes-to-each-guiders-tap-schedule-so-that-they-are-based-on-existing-operatio)

PR #652, opened 2019-10-27
Improve layout of navigation bar

benlovell:
This would wrap for users with the administrator role, since the
provider dropdown was styled incorrectly and has recently increased in
width. The app title has changed to drop the 'Appointment', this matches
the other planners that also changed recently and has the useful effect
of freeing up some valuable horizontal real estate.


benlovell:
[Improve wrapping navigation](https://trello.com/c/3jGeM6Ig/1175-improve-wrapping-navigation)

PR #654, opened 2019-11-05
Only enforce notes after cut off date

benlovell:
Changes logic to only enforce the presence of notes after the initial
change was introduced. This is to permit changes to appointments
pre-dating that time.


benlovell:
[Amend validation so that TAP users can update appointment statuses without inputting into additional information box](https://trello.com/c/4kGawVSM/1182-amend-validation-so-that-tap-users-can-update-appointment-statuses-without-inputting-into-additional-information-box)

PR #659, opened 2019-12-10
Reduce grace period to effectively one day

benlovell:
Reduces the grace period to one day across all providers per a request
from TPAS.


benlovell:
[Decrease TPAS grace period to 24 hours](https://trello.com/c/3BH41uDA/1222-decrease-tpas-grace-period-to-24-hours)

PR #662, opened 2020-01-12
Ensure Notify errors are correctly instrumented

benlovell:
Vital information was being lost when these errors were reported to
bugsnag. A recent change to the Notify client ensures the errors are
constructed correctly.


benlovell:
[Instrument Notify errors correctly](https://trello.com/c/33uYDFCL/1254-instrument-notify-errors-correctly)

PR #663, opened 2020-01-13
Ensure SMS candidates are correctly determined

benlovell:
We should ensure that 'mobile' numbers are 'valid' UK mobile numbers
before including them in the SMS reminder candidate list.

Prior to this change there was no checks performed against the mobile
field and regular numbers as well as international numbers were creeping
through.


benlovell:
[Identify SMS reminder candidates correctly](https://trello.com/c/ghJE7Gc1/1255-identify-sms-reminder-candidates-correctly)

PR #665, opened 2020-01-21
Validate length of phone/mobile for TP agents

benlovell:
We've noticed an increase in invalid phone numbers (usually lacking
digits) recently when entered by agents during the booking process. This
change ensures a phone or mobile number (when supplied) will be at least
11 digits in length.


benlovell:
[Validate phone numbers](https://trello.com/c/OsLjW7Gp/1263-validate-phone-numbers)

PR #674, opened 2020-03-05
Don't trigger audits after redacting

benlovell:
Previously this would trigger the creation of an activity that would
then be enqueued to pusher, meanwhile the activities are destroyed as
part of the redaction process and once the job is dequeued, the
underlying activity is gone so would cause an error.


benlovell:
[Ensure redactions don't trigger realtime notifications](https://trello.com/c/4z2XdCW6/1331-ensure-redactions-dont-trigger-realtime-notifications)

PR #676, opened 2020-03-10
Ignore bounced accessibility adjustment emails

benlovell:
These were firing web hooks but are not handled internally by resource
managers, plus they appear as bounce activities against the associated
booking, thus confusing agents that may be viewing the activity feed. We
can just ignore these particular calls instead.


benlovell:
[Ignore bounced accessibility adjustment emails for resource managers](https://trello.com/c/t4faNgfV/1340-ignore-bounced-accessibility-adjustment-emails-for-resource-managers)

PR #682, opened 2020-03-23
Add the first wave of CITA providers

benlovell:
This will aid the channel shift from face-to-face to telephone whilst the
ongoing health crisis unfolds.


benlovell:
[Set up F2F delivery centres as TAP providers](https://trello.com/c/GW5ZsPCO/1355-set-up-f2f-delivery-centres-as-tap-providers)

PR #685, opened 2020-03-23
Reverse the ordering of guider schedules

benlovell:
This was requested by resource managers to aid the location of the most recent
schedule for a particular guider.


benlovell:
[As a Resource Manager, I need to see schedules shown in descending order so that I can set guider availability more efficiently](https://trello.com/c/2RG5xeJ0/1003-as-a-resource-manager-i-need-to-see-schedules-shown-in-descending-order-so-that-i-can-set-guider-availability-more-efficiently)

PR #686, opened 2020-03-24
Revert "Allow for timezone change"

benlovell:
This reverts commit 1c039b2b08bec484733d9ea5b222411ef97aa308.


benlovell:
[As a customer who has booked an appointment I need to be aware that an appointment time is in DST/BST (2020)](https://trello.com/c/ZkzLaWpF/1356-as-a-customer-who-has-booked-an-appointment-i-need-to-be-aware-that-an-appointment-time-is-in-dst-bst-2020)

PR #687, opened 2020-03-25
Resource manager email notifications

benlovell:
Notify resource managers by email when an appointment has changed. Does not
apply to appointments pertaining to the TPAS organisation as they receive
realtime notifications.


benlovell:
[As TAP Resource Manager I need to know when phone appointments are booked, cancelled, rescheduled or rebooked so I can update the delivery partner CRM](https://trello.com/c/jMTwUD01/1085-as-tap-resource-manager-i-need-to-know-when-phone-appointments-are-booked-cancelled-rescheduled-or-rebooked-so-i-can-update-the)

PR #688, opened 2020-03-26
Allow non-TPAS users to 'process' appointments

benlovell:
This marks an appointment as synchronised with casebook and only applies to
non-TPAS providers.


benlovell:
[As a resource manager at CITA delivery centres, I want to know when telephone appointments are processed on TAP](https://trello.com/c/jQGWnXeH/1359-as-a-resource-manager-at-cita-delivery-centres-i-want-to-know-when-telephone-appointments-are-processed-on-tap)

PR #690, opened 2020-03-31
Improve the redactor

benlovell:
Can redact a single appointment instance by `REFERENCE` or can now
redact all records older than 2 years for GDPR purposes.


benlovell:
[Ensure all customer data is automatically removed from booking systems and Zendesk from 2020-21](https://trello.com/c/GD7b42b2/1303-ensure-all-customer-data-is-automatically-removed-from-booking-systems-and-zendesk-from-2020-21)

PR #691, opened 2020-03-31
Ensure bookable slots are also wiped

benlovell:


benlovell:
[Wipe schedules, slots and holidays across all providers](https://trello.com/c/DHiFFNCr/1366-wipe-schedules-slots-and-holidays-across-all-providers)

PR #692, opened 2020-03-31
Don't email alert NI resource managers

benlovell:
They don't find these alerts useful.


benlovell:
[Turn off TAP notifications for Pension Wise Northern Ireland](https://trello.com/c/Ww6oBjBy/1370-turn-off-tap-notifications-for-pension-wise-northern-ireland)

PR #693, opened 2020-04-01
Ensure allocation does not error

benlovell:
This was previously erroring when the resource manager did not provide
an adhoc start date/time.


benlovell:
[Ensure adhoc allocation doesn't error when a resource manager forgets to specify an appointment date/time](https://trello.com/c/XUgPciiI/1375-ensure-adhoc-allocation-doesnt-error-when-a-resource-manager-forgets-to-specify-an-appointment-date-time)

PR #694, opened 2020-04-02
Add a filter for `processed` appointments

benlovell:
Defaults to all appointments for TPAS and shows filters for other
organisations. Defaults to 'no' for the organisations it applies to.


benlovell:
[Add mark as processed filter to TAP](https://trello.com/c/JigsjiA7/1377-add-mark-as-processed-filter-to-tap)

PR #695, opened 2020-04-02
Temporarily exclude CAS slots from API respinse

benlovell:
This will ensure CAS can add slots that won't be immediately filled by
customer demand.


benlovell:
[Temporarily exclude CAS slots from customer-facing API](https://trello.com/c/esgOLk3y/1380-temporarily-exclude-cas-slots-from-customer-facing-api)

PR #696, opened 2020-04-03
Add a temporary task to reallocate appointments

benlovell:
This is a rough spike at a task we can use to urgently move allocated
appointments across providers.


benlovell:
[Move TPAS appointments to a different TAP organisation](https://trello.com/c/oxXVnI7g/1376-move-tpas-appointments-to-a-different-tap-organisation)

PR #699, opened 2020-04-06
Configure bank holidays up until May 2020

benlovell:


benlovell:
[Configure bank holidays for 2020](https://trello.com/c/1hfE68kk/1383-configure-bank-holidays-for-2020)

PR #700, opened 2020-04-06
Don't push realtime to guider activity feed

benlovell:
The guider's have their own realtime feed for activity pertaining to
their appointments. Since they always view the activity through the
perspective of a particular appointment, the overhead involved in
sending and handling these notifications isn't worth the effort.

I also realised a point where duplicate push notifications were being
sent.


benlovell:
[Investigate pusher spike](https://trello.com/c/IOoFW6FB/1385-investigate-pusher-spike)

PR #702, opened 2020-04-14
Accommodate timezones for grace period

benlovell:
Previously this was allowing appointments to be created on the same,
given business day, as the timezone offset was not being applied.


benlovell:
[Respect DST/BST whilst determining the earliest available slot](https://trello.com/c/uKdZkXnV/1392-respect-dst-bst-whilst-determining-the-earliest-available-slot)

PR #704, opened 2020-04-15
Exclude bank holidays for CAS

benlovell:
Assigns holidays for all other users.


benlovell:
[Revisit bank holidays for August onward](https://trello.com/c/addoYIKy/1384-revisit-bank-holidays-for-august-onward)

PR #707, opened 2020-04-17
Display a banner for DC pension unsure customers

benlovell:
When the customer is unsure they have a DC pension we display a banner
to alert anyone viewing the appointment.


benlovell:
[Display customer DC pot responses on TAP appointment records](https://trello.com/c/7K7uzfiG/1403-display-customer-dc-pot-responses-on-tap-appointment-records)

PR #709, opened 2020-04-20
Reduce grace period to one working day

benlovell:


benlovell:
[Adjust grace period for TPAS](https://trello.com/c/A0NppNLe/1406-adjust-grace-period-for-tpas)

PR #710, opened 2020-04-22
Extend close-of-play to 9pm

benlovell:
Effectively set the CoP to 9pm. This was 6:30pm previously and would
cause confusion among some delivery providers.


benlovell:
[Change end of business day](https://trello.com/c/qTTlarPN/1412-change-end-of-business-day)

PR #711, opened 2020-04-26
Bump TPAS-specific grace period to 48 hours

benlovell:
Per request from TPAS.


benlovell:
[Adjust grace period for TPAS](https://trello.com/c/A0NppNLe/1406-adjust-grace-period-for-tpas)

PR #712, opened 2020-04-28
Return a single result regardless if processed

benlovell:
Directs to a single result regardless whether the appointment was
processed.


benlovell:
[Ensure that search button always takes you to full search page when there's only one unprocessed appointment](https://trello.com/c/bLGfyybD/1405-ensure-that-search-button-always-takes-you-to-full-search-page-when-theres-only-one-unprocessed-appointment)

PR #713, opened 2020-05-04
Allow Business Analysts to set organisation

benlovell:
Allows CITA (or other) business analysts to switch their
provider/organisation.


benlovell:
[As a Citizens Advice business analyst I should be able to view TAP appointments for all CITA TAP organisations](https://trello.com/c/y1VEMiQd/1421-as-a-citizens-advice-business-analyst-i-should-be-able-to-view-tap-appointments-for-all-cita-tap-organisations)

PR #714, opened 2020-05-07
Removing this holiday to shift CAS slots

benlovell:
CAS have a handful of slots free on Saturday that would otherwise have
been skipped by the grace period.


benlovell:
[Ensure CAS slots are bookable for Sat 9 May (Bank Holiday Weekend)](https://trello.com/c/z2LLiTGV/1427-ensure-cas-slots-are-bookable-for-sat-9-may-bank-holiday-weekend)

PR #719, opened 2020-06-08
Third party bookings

benlovell:
Allows agents and others to specify a third party will take the
appointment on behalf of the given customer. When this is specified, the
agent will also be asked to confirm the customer gave their consent.
They will also see a templated area in the notes field to provide
further information.

Resource managers will receive email notification when the third party
option is specified. This is now combined with what was previously the
accessibility adjustments email. A banner will also be displayed when
editing/viewing the appointment.


benlovell:
[As a telephone guider I need to be aware if a Pension Wise appointment is being delivered to a third party and record consent where appropriate](https://trello.com/c/1uMhRYyl/1446-as-a-telephone-guider-i-need-to-be-aware-if-a-pension-wise-appointment-is-being-delivered-to-a-third-party-and-record-consent-wh)

PR #723, opened 2020-06-25
Log a 'drop activity' for invalid emails

benlovell:
When the email attempted causes an SMTP error we ought to log an
activity so the agent or other owner gets an opportunity to correct the
underlying issue.


benlovell:
[Add additional bounce activity for TAP feed](https://trello.com/c/DmAXRbmP/1464-add-additional-bounce-activity-for-tap-feed)

PR #968, opened 2024-07-18
Broaden duplicates to non-pending too

benlovell:
This will keep the notifications to pending-only but broaden the display of the duplicates during appointment viewings.


benlovell:
Closing in favour of #962.

PR #1056, opened 2025-07-04
Add the basis for online rescheduling

benlovell:
Adds two primary APIs. One for matching up an appointment that's
eligible for rescheduling, given the appointment reference and
customer's date of birth. The other is for the rescheduling process
itself. When the guider allocation switches organisations, we notify
both sets of resource managers.

Casebook rescheduling is handled in the simplest case, whereby both
guiders (if reallocating cross-organisation) are switched automatically.


benlovell:
Closing in favour of #1060.

